PR #16, opened 2016-05-18
Preparation for publishing Mendel 1.0.0; No more libs, 3 new packages

irae:
This is probably the last large breaking change we will introduce.

We need to push to NPM in order to use internally at Yahoo, so I dealt with all the symbolic linking problems. Now we have a lot of smaller packages and linkall takes more time to run, but it also runs on the examples and runs on the root of the project.

Here is a summary of changes:

  * New mendel-config package: only config and variation parsing. We can in the future improve and have only index.js in this repo, merging variations in the config
  * New mendel-core package: This is all the production level shared dependencies. I am not calling it "production" to avoid confusion. In the future we might merge the loader into mendel-core instead of being a dependency.
  * New mendel-development package: all the development shared dependencies
  * Removed 1 old package, lib folder and all symbolic links that were on git
  * Removed "private:true" from package.json so we can finally push to npm
  * Linkall now runs in many folders and honors package name. It will loop through all packages, through all examples and also on the root of the project and crosslink and install everything.

Everything else is refactoring to require the right files, and tests to pass. There is one small fix that the loader required relative paths and the full-example was broken.

@gomezd


irae:
Forgot to say. It is probably easier to pull this changes and test than reviewing all the code. The actual diff without the moving files is the second commit, it is easier to read the changes there. It is almost all "require(....)" changes.

packages/mendel-development-middleware/package.json
```diff
@@ -18,11 +18,14 @@
...
+    "mendel-config": "1.0.0",
+    "mendel-treenherit": "1.0.0",
+    "mendel-loader-server": "1.0.0",
+    "mendel-requirify": "1.0.0",
+    "mendel-development": "1.0.0"
   },
   "peerDependencies": {
-    "express": "*"
+    "express": "*",
+    "path-to-regexp": "*"
```
irae: This was important, this way we don't require our own version of path-to-regexp.

examples/full-example/app.js
```diff
@@ -19,7 +19,7 @@ app.get('/', function(req, res) {
 
     if (serverRender) {
         var resolver = req.mendel.resolver(variations);
-        var Main = resolver.require('main.js');
+        var Main = resolver.require('./main.js');
```
irae: This was broken already. Just fixing it. @gomezd 😉

gomezd:
It's looking good, only 2 things:
- `linkall` does not work in a clean repo (`mendel-config` has to be linked first)
- Do we want to keep shorter name `mendel-middleware` for production?

Following this I can split `mendel-loader-server` for production and `mendel-loader-server-development` (?) too long!

gomezd:
[LGTM](http://lgtm.in/i/Qegg5iK6q)

irae:
About development and production for the loader, yes, you can split it if you want. Or you can merge it on each respective middleware. I am not sure about the shorter name. Let's sit on this for a while and we make a call later.

PR #18, opened 2016-05-18
Run `mendel-requiriy` with mendel bin

gomezd:
When running `mendel` bin, server side files should be generated if `serveroutdir` config is present.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,9 @@
     "glob": "^7.0.3",
     "mkdirp": "^0.5.1",
     "xtend": "^4.0.1",
-    "mendel-config": "1.0.0"
+    "mendel-config": "1.0.0",
+    "mendel-browserify": "1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -64,6 +64,8 @@
     "rimraf": "^2.5.2",
     "tap": "^5.7.0",
     "temp": "^0.8.3",
-    "through2": "^2.0.1"
+    "through2": "^2.0.1",
+    "mendel-browserify": "1.0.0",
+    "mendel-requirify": "1.0.0"
```
irae: I don't think those are `devDependencies`, right? When someone "npm install mendel -g" we want those to be there, right?

irae:
[LGTM](http://lgtm.in/i/s2g8nrFQp)

[:+1:](http://lgtm.in/u/s2g8nrFQp)

PR #19, opened 2016-05-19
Added swatch worker to run server side dev watch on a child process.

gomezd:
This adds performance improvements:
- Running all the server side watchify pipelines in a child process.
- Sharing watchify cache between bundlers cuts time in almost half.

### DIFF
--- a/packages/mendel-development-middleware/index.js
+++ b/packages/mendel-development-middleware/index.js
@@ -12,7 +12,7 @@ var treenherit = require('mendel-treenherit');
 var parseConfig = require('mendel-config');
 var validVariations = require('mendel-config/variations');
 var resolveVariations = require('mendel-development/resolve-variations');
-var Swatch = require('./swatch');
+var swatch = require('./swatch-worker');
 var CachedStreamCollection = require('./cached-stream-collection');
 var MendelLoader = require('mendel-development-loader');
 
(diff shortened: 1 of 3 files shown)


irae:
I am trying this on our private application (my branch is rebased to master) and it still takes a lot of time for the CPU to be at rest at startup.

Also, I have a "button_color" variation. It is originally "red" and when I change to Yellow and refresh the page, I aways get the server-side portion in red, although the client bundle quickly catches up.

The lifecycle is like this using `watch:false` on the MendelMiddleware:
  * Startup super quick, in about 10 seconds server is listening and idle in Activity Monitor
  * Loading our base variation with &ssr=false about 1 second but is a blank page
  * Takes the vendor and product_app buindles takes 17 and 26 seconds to compile respectivelly
  * Change URL to &variations=button_color and get same results for bundles but 300ms for server route
  * Save button color: takes 2 seconds to serve each bundle
  * Reload: about 100 ms each bundle
  * Save and reload cycle goes down to 300ms each save and 100ms each repeated serve.

Now removing the `watch:false` with this branch:
  * Server takes 1 minute 15 seconds to become idle
  * same 1 second to render page with SSR on
  * bundles took 19 and 26 seconds to render, it's the same
  * Change URL to &variations=button_color and same results as above
  * The other timings are ok for bundles, same results.
  * But for some reason, the second color I change sticks on the server side forever. If I start with pink, save the file as yellow, and than blue, it keeps yellow, and I can change many times and never changes from yellow. Client bundles change consistently.

So, besides the startup issue persisting, this is a real improvement over the 3x it takes to do this on the same process.

Good Job!

irae:
[LGTM](http://lgtm.in/i/s4l0wulWp)

[:+1:](http://lgtm.in/u/s4l0wulWp)

PR #22, opened 2016-05-23
Block incomming requests while requirify writes server side files

gomezd:
- Added `writeCache` to `mendel-requirify` for incremental changes
- Improved `swatch` events and timing
- Fixed tests

With this change the dev middleware will block requests on app start until server side files are built, and on source file changes, attempting to avoid race conditions on incoming requests while reloading.

### DIFF
--- a/test/trees.js
+++ b/test/trees.js
@@ -23,11 +23,11 @@ test('MendelTrees initialization', function (t) {
 
     process.chdir(appPath);
 
-    if(fs.existsSync(manifestPath)) fs.unlinkSync(manifestPath);
-    t.throws(MendelTrees, /bundle at path/, 'requires manifest to exist');
-
     fs.writeFileSync(manifestPath, '{invalid json}');
-    t.throws(MendelTrees, /Invalid bundle file/, 'requires valid manifest');
+    t.throws(MendelTrees, {message: /Invalid bundle file/}, 'requires valid manifest');
+
+    if(fs.existsSync(manifestPath)) fs.unlinkSync(manifestPath);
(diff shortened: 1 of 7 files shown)


test/trees.js
```diff
@@ -23,11 +23,11 @@ test('MendelTrees initialization', function (t) {
...
 
-    if(fs.existsSync(manifestPath)) fs.unlinkSync(manifestPath);
-    t.throws(MendelTrees, /bundle at path/, 'requires manifest to exist');
-
     fs.writeFileSync(manifestPath, '{invalid json}');
-    t.throws(MendelTrees, /Invalid bundle file/, 'requires valid manifest');
+    t.throws(MendelTrees, {message: /Invalid bundle file/}, 'requires valid manifest');
+
+    if(fs.existsSync(manifestPath)) fs.unlinkSync(manifestPath);
+    t.throws(MendelTrees, {message: /Could not find/}, 'requires manifest to exist');
```
irae: Thanks for this fix. I looked a bit into it Friday and got back to testing only on node 0.12. With this fix it works on both Node 6 and Node 0.12 🎆

irae:
Code looks ok, but I am getting consistent error thrown once the request is ready after an update. Also, sometimes I save a file ant it never changes. Looks like erasing all the files in the build folder before starting the server helps having a more consistent behavior for whatever reason.

I will send you a private message on how to reproduce the errors.

About making it server render only route, you can do it if you lazy initialize on "req.mendel.resolver()" but I think it is better to be a separate PR.

gomezd:
I'm closing this, since the server side watch will most likely be re architected.

PR #23, opened 2016-05-24
Make sure we honor externals, excluded and ignores in all pipelines

irae:
During development, we were not preventing external bundles to be included. This means that in the example/full-example we had a main app with react, in addition to react in the vendor file. Production was never affected.

This fixes it by using the same option in all places we create bundles ourselves.

Tests will follow on a separate PR because we have no unit tests for the middleware so far.


gomezd:
👍 

LGTM

PR #25, opened 2016-05-24
Merge environment config overrides

gomezd:
Added the option to override configurations based on `MENDEL_ENV` or `NODE_ENV` environment variables.

Config options can be overriden with the `env` config value in `.mendelrc`, `package.json` or through the config api params. 

```yaml
base: default
basetree: src/default

bundles:
  foo:
    entries:
      - index.js

variations:
  variation_A:
  variation_B:

env:
  production:
    bundles:
       foo:
          transform:
            - uglify
```
Values are merged recursively

### DIFF
--- a/packages/mendel-config/config.js
+++ b/packages/mendel-config/config.js
@@ -7,54 +7,55 @@ var path = require("path");
 var fs = require("fs");
 var xtend = require("xtend");
 var yaml = require('js-yaml');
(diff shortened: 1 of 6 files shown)


irae:
[LGTM](http://lgtm.in/i/Q2iqOXgWp)

[:+1:](http://lgtm.in/u/Q2iqOXgWp)

PR #26, opened 2016-05-26
Pass variations with base to `mendelify` transform.

gomezd:
`mendelify` transform needs to know about base variation to fall back when no variations exist.
This is needed because `mendel-requirify` stream transform checks for `row.variation` to be present so it can write the output file.


tufandevrim:
👍

irae:
👍

PR #27, opened 2016-05-26
Fix mendel plugin recursion when using programmatically

irae:
```js
var fs = require('fs');
var babelify = require('babelify');
var browserify = require('browserify');
var mendelPlugin = require('mendel-browserify');

var b = browserify({
    entries: ['./isomorphic/base/main.js'],
    transform: [babelify],
    plugin: [mendelPlugin]
});
b.external(['react', 'react-dom']);

b.bundle().pipe(fs.createWriteStream('main.programmatic.js'))
```

The code above is valid and would go with infinite recursion. PR adds a filter on the plugin and add tests.

@tufandevrim pay attention to the `constructor` part because I think you'll need to add this to extractify

### DIFF
--- a/packages/mendel-browserify/index.js
+++ b/packages/mendel-browserify/index.js
@@ -60,6 +60,17 @@ function MendelBrowserify(baseBundle, opts) {
         var vopts = xtend(self.baseOptions);
         var browserify = baseBundle.constructor;
         var pipeline = baseBundle.pipeline.constructor;
+
+        vopts.plugin = [].concat(vopts.plugin).filter(Boolean)
+            .filter(function(plugin) {
+                if (typeof plugin === 'string') {
+                    return plugin !== 'mendel-browserify';
+                } else if(MendelBrowserify.constructor === plugin.constructor) {
+                    return false;
+                }
+                return true;
+            });
+
         var variationBundle = browserify(vopts);
 
         self.prepareBundle(variationBundle, variation);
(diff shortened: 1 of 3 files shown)


packages/mendel-browserify/index.js
```diff
@@ -60,6 +60,17 @@ function MendelBrowserify(baseBundle, opts) {
...
         var pipeline = baseBundle.pipeline.constructor;
+
+        vopts.plugin = [].concat(vopts.plugin).filter(Boolean)
+            .filter(function(plugin) {
+                if (typeof plugin === 'string') {
+                    return plugin !== 'mendel-browserify';
+                } else if(MendelBrowserify.constructor === plugin.constructor) {
+                    return false;
+                }
+                return true;
```
gomezd: There is a third case where plugin is an array of `[function, options]`

tufandevrim:
:+1:

packages/mendel-browserify/index.js
irae: Will fix that before merging! Good catch!

packages/mendel-browserify/index.js
```diff
@@ -7,6 +7,7 @@ var xtend = require('xtend');
 var defined = require('defined');
 var fs = require('fs');
 var path = require('path');
+var isarray = require('isarray');
```
tufandevrim: why not use Array.isArray?

irae: Mendel codebase is ES5 safe, the tests will run in NodeJS 0.12 because I want teams at Yahoo that are on 0.12 to use Mendel.

PR #28, opened 2016-05-26
Remove all uses of isarray

irae:
@tufandevrim thanks! I totally missed that we have it in all node versions and browsers we support


tufandevrim:
:+1:

PR #29, opened 2016-05-26
Check for `serveroutdir` on dev loader

gomezd:
Added check `loader.isSsrReady()` or with the middleware enhanced request `req.mendel.isSsrReady()`

### DIFF
--- a/packages/mendel-development-loader/loader.js
+++ b/packages/mendel-development-loader/loader.js
@@ -2,6 +2,7 @@
    Copyrights licensed under the MIT License.
    See the accompanying LICENSE file for terms. */
 
+var fs = require('fs');
 var MendelResolverDev = require('./resolver');
 var resolveVariations = require('mendel-development/resolve-variations');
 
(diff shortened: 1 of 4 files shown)


packages/mendel-development-loader/loader.js
```diff
@@ -13,11 +15,45 @@ function MendelLoaderDev(existingVariations, config, parentModule) {
...
+    return new MendelResolverDev(this._parentModule, dirs, config);
+};
+
+MendelLoaderDev.prototype.isSsrReady = function() {
+    return this._ssrReady;
+};
+
+MendelLoaderDev.prototype._fallback_resolve = function (mod) {
+    var config = this._config;
+    return path.resolve(config.basedir, config.basetree, mod);
```
irae: I see how this might help, but I would prefer to render nothing than rendering the source of the base tree.

If babel register is not being used, it would just throw. Also, if the user ins on and experiment on the client bundle, there will be mismatch of markup and rehydration would fail.

Can you help me better understand how this is helping?

gomezd: Good point. I'll change it to a warning and let it throw.

irae:
👍

PR #30, opened 2016-06-02
Add lazy button into the example

tufandevrim:
Adding lazy button the test to test extractify plugin in the future

### DIFF
--- a/examples/full-example/isomorphic/base/components/toolbar.js
+++ b/examples/full-example/isomorphic/base/components/toolbar.js
@@ -1,14 +1,39 @@
 import React from 'react';
 import Button from './button';
 
+let Lazy;
+
 class Toolbar extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            laoded: false
+        }
+        this.handleClick = this.handleClick.bind(this)
+    }
+
+    handleClick() {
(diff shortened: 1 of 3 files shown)


examples/full-example/isomorphic/base/components/toolbar.js
```diff
@@ -1,14 +1,39 @@
...
     render() {
         return (
-            <nav className="toolbar">
-                <Button />
-                <Button />
-                <Button />
-            </nav>
+            <div>
+                <nav className="toolbar">
+                    <Button onClick={this.handleClick}>Button #1</Button>
```
stephanwlee: don't you have to `bind` in es6 class inherited react? :P

tufandevrim: Check the line 12

gomezd:
👍 

LGTM

examples/full-example/isomorphic/base/components/toolbar.js
irae: Line 12 looks ugly, but since we don't use classes I am not sure if this is a good or a bad idea. It actually does the job well, so I think this is ok for now.

irae:
LMAO that image

PR #31, opened 2016-06-03
Removed swatch in favor of watchified bundles in dev middleware.

gomezd:
- Removed swatch and added `mendel-requirify` to the dev middleware bundlers to generate server side files on demand without significant overhead.
- Loader detects if `serveroutdir` exists, it can be queried through `isSsrReady` method, also available in the middleware through `req.mendel.isSsrReady`.
- Improved `mendel-requirify` so that it doesn't add `__mendel_require_()` to `node_modules` dependencies.
- Improved dev middleware watchified bundles speed by re-bundling on update, which handles client and server side.
- Fixed tests


packages/mendel-development-loader/loader.js
```diff
@@ -14,7 +14,7 @@ function MendelLoaderDev(existingVariations, config, parentModule) {
     this._existingVariations = existingVariations;
     this._config = config;
     this._parentModule = parentModule || module.parent;
-    this._ssrReady = isValidDir(config.serveroutdir);
```
irae: How about instantiate as `false` first? Looks like the new function can return undefined for a while?

packages/mendel-requirify/index.js
```diff
@@ -33,10 +28,23 @@ function requirify(b, opts) {
...
-            out.end(done);
+            var match;
+            dirs.some(function(dir) {
+                var parts = file.split('/' + dir + '/');
+                var found = parts.length > 1;
+                if (found) {
+                    match = [outdir, dir, parts[1]];
+                }
+                return found;
+            });
```
irae: This looks very much with `variationMatches`. In case it is the same you can re-use the function.

packages/mendel-development/require-transform.js
```diff
@@ -14,15 +14,24 @@ function _wrap(src) {
...
     var _src = falafel(src, opts, function (node) {
         if (isRequire(node)) {
             var module = node.arguments[0].value;
-            node.update("__mendel_require__('" + module + "')");
+
+            var match;
+            dirs.some(function(dir) {
+                match = module.split('/' + dir + '/')[1];
+                return !!(match);
+            });
```
irae: Again similar to `variationMatches`. Did you / could you double check all those cases and see if they can share helper functions?

irae:
Code-wise this looks good. I will test a bit once I get a chance.

Architecture-wise I am not very found of the approach, since this means we have to start in development mode without any application code and wait for the browser to request the javascript file in order to have some code to server side render.

It also have a pitfall, if an application uses `package.json` with "browser" and "main" entries, you will get the browser version compiled as a server-file.

I think this is better than what we have and will enable us to move forward. But we are still without a final solution for server-side-development.

Let's test this on the internal projects and talk more in person.

👍

packages/mendel-development-loader/loader.js
gomezd: Done.

gomezd:
Fixed comments and example app.

PR #32, opened 2016-06-03
Quality of life manifest generation improvements

irae:
When there is a sha colision, we how write the files to a temp folder
so it is easier to debug.

Also, sorting the manifest makes it easier to diff previous versions
with new versions of the application. This is good for debugging and
could be good for diagnose real production issues on the applications
using mendel, in case they store all manifests with the application
package.


packages/mendel-browserify/index.js
```diff
@@ -203,6 +210,35 @@ MendelBrowserify.prototype.doneManifest = function() {
...
+        indexes: {},
+        bundles: []
+    };
+
+    var self = this;
+    Object.keys(this._manifestIndexes).sort().forEach(function(file) {
+        var bundle = self._manifestBundles[self._manifestIndexes[file]];
+        sortedManifest.bundles.push(bundle);
+
+        var index = sortedManifest.bundles.indexOf(bundle);
```
gomezd: index = sortedManifest.bundles.length - 1 ?

irae: WAT? It is working, if it was wrong it would totally break the tests and examples. Besides, up on the file there is a very similar line and it is the same for ages.

gomezd: I didn't mean it was wrong :) shoul have been more explicit, i meant that since you just push to the array, then the index is `sortedManifest.bundles.length - 1` right? And you wouldn't need to search the array with indexOf and could arguably be faster...

gomezd:
👍 
[LGTM](http://lgtm.in/i/YgPMcYDNp)

packages/mendel-browserify/index.js
irae: Oh! Sorry, I probably misread your line when I first read.

Yeah, I could do that too. And could also have used the second argument of the forEach funcion for index. Since I have a similar code during manifest generation — and there we have potential racing conditions — I chose to make it look like the other line.

There is a bit code swag there 😄:
It feels natural "add the bundle to the array, retrieve the bundle position, and store the index under the file key" feels better than "add the bundle to the array, get the position for the last item in the array, , and store the index under the file key".

PR #33, opened 2016-06-10
Fixing mendel manifest generation to work better with exposed modules and JSON

irae:
Handling of exposed modules only worked correctly if external libraries didn't depend on each other. The full example still needs to be updated to cover this case, which would happen if some other library required React on vendor library. Also, if we require a file (instead of using entries)  that is inside base folder, it would also need this patch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mendel",
-  "version": "1.0.5",
+  "version": "1.0.6",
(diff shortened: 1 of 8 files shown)


irae:
@gomezd can you take a look when you have the time?

gomezd:
👍 
[LGTM](http://lgtm.in/i/AQD2Mzm5q)

PR #34, opened 2016-08-09
[DNM] Adding getExposed method

tufandevrim:
@irae here is what I come up with for getting exposed dependencies.
If this is fine, I'll add same function into `mendel-development-middleware` and some tests.

Let me know what you think

### TODO:
* [x] update the example to load lazy component on demand
* [ ] update `mendel-development-middleware`
* [ ] rebase

### DIFF
--- a/examples/full-example/app.js
+++ b/examples/full-example/app.js
@@ -24,9 +24,21 @@ app.get('/', function(req, res) {
         optionalMarkup = ReactDOMServer.renderToString(Main())
     }
 
+    var exposed = req.mendel.getExposed(variations);
+    var depBundleMap = Object.keys(exposed).reduce(function(acc, bundle) {
+        exposed[bundle].deps.forEach(function(dep) {
+            acc[dep.expose] = exposed[bundle].path;
+        });
+        return acc;
+    }, {});
+
     var html = [
         '<!DOCTYPE html>',
-        '<html><head></head><body>',
+        '<html><head>',
+       '<script>window.exposedDeps=',
+        JSON.stringify(depBundleMap),
+        ';</script>',
+        '</head><body>',
             '<div id="main">'+optionalMarkup+'</div>',
             bundle(req, 'vendor', variations),
             bundle(req, 'main', variations),
--- a/packages/mendel-middleware/index.js
+++ b/packages/mendel-middleware/index.js
@@ -20,6 +20,7 @@ function MendelMiddleware(opts) {
     var loader = new MendelLoader(trees, {
         parentModule: module.parent
     });
+    var exposedBundleVariations = {};
 
     return function(req, res, next) {
         req.mendel = req.mendel || {};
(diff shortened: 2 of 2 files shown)


packages/mendel-middleware/index.js
```diff
@@ -29,6 +30,38 @@ function MendelMiddleware(opts) {
...
+            if (!variations) {
+                variations = 'base';
+            }
+
+            if (exposedBundleVariations[variations]) {
+                return exposedBundleVariations[variations];
+            }
+
+            var exposedBundleVariation = Object.keys(trees.bundles).reduce(function(acc, bundle) {
+                var vTree = trees.findTreeForVariations(bundle, variations);
```
irae: This is the bottleneck for performance. We can't call this on every bundle on every request. Also, we don't need variations for that.

examples/full-example/app.js
```diff
@@ -24,9 +24,21 @@ app.get('/', function(req, res) {
         optionalMarkup = ReactDOMServer.renderToString(Main())
     }
 
+    var exposed = req.mendel.getExposed(variations);
```
irae: This calls `trees.findTreeForVariations` 3 times (one for each current bundle.

examples/full-example/app.js
```diff
@@ -24,9 +24,21 @@ app.get('/', function(req, res) {
...
         '<!DOCTYPE html>',
-        '<html><head></head><body>',
+        '<html><head>',
+       '<script>window.exposedDeps=',
+        JSON.stringify(depBundleMap),
+        ';</script>',
+        '</head><body>',
             '<div id="main">'+optionalMarkup+'</div>',
             bundle(req, 'vendor', variations),
             bundle(req, 'main', variations),
```
irae: Than we call `trees.findTreeForVariations` twice here. It's a 5 times total, where it should be 3 at max.

packages/mendel-middleware/index.js
```diff
@@ -29,6 +30,38 @@ function MendelMiddleware(opts) {
             return getPath({ bundle: bundle, hash: tree.hash });
         };
 
+        req.mendel.getExposed = function(variations) {
+            if (!variations) {
+                variations = 'base';
+            }
+
+            if (exposedBundleVariations[variations]) {
```
irae: There is a number of reasons you can't cache like this:

1. This variations are an array, so each request is a new reference, so this is a memory leak.
2. `['base', 'bucket_A']` is the same as `['bucket_A', 'base']` and the logic to normalize, sanitize and order by configuration precedence is inside `trees` so you the proper place to cache would be there.
3. Even if you sanitize, sort and join the variations, this is a vector for a DDOS attack because you can create infinite keys (although the objects would be the same).
4. In multi tenant deployments (one process per processor, between 4 and 16 processors per server and many servers per datacenter, like yahoo has) you will store cache in each process and consume a lot of memory.

I think these are all the reasons we don't cache in Mendel middleware or in `MendelTrees`

packages/mendel-middleware/index.js
tufandevrim: Agreed this could be performance bottleneck for the first request of a specific variation. After the first request, it should be served form `exposedBundleVariations` object as a cache.

examples/full-example/app.js
tufandevrim: We could cache the result on first 2 as well.

packages/mendel-middleware/index.js
tufandevrim: Yeah I see the mistake. I assumed that `variations` would be string but it is an array. 

All your concerns are valid.

irae:
Besides all the comments about caching and all, relying on variations is a bad idea for the development middleware too. It will be much easier to create the development version with just figuring out the entries.

I'll extract the app changes from my branch and make a PR to master to  be able to test lazy in all variations, this might be helpful to test both proposals for middleware.

tufandevrim:
Closing it in favor of #35

PR #35, opened 2016-08-10
Alternative draft for exposing bundles

irae:
The idea of this PR is to demonstrate a way to give enough information to the applications using mendel, without changing the     performance characteristics we have. We are avoiding at all cost calling `findTreeForVariations` without real need.

If we ever have more than one method that needs it, we will need caching per request or maybe change to a stateful API.

@tufandevrim this is a counter-proposal for #34

### DIFF
--- a/examples/full-example/app.js
+++ b/examples/full-example/app.js
@@ -33,7 +33,7 @@ app.get('/', function(req, res) {
             // WARNING: This temporary "lazy" bundle is not part of the example
             // The dev team needs this here for a while in order to
             // do some work in parallel branches
-            bundle(req, 'temporary', variations),
+            entryMap(req, 'temporary', variations),
             // END WARNING
         '</body></html>'
     ].join('\n');
@@ -45,6 +45,29 @@ app.get('/', function(req, res) {
 function bundle(req, bundle, variations) {
     var url = req.mendel.getURL(bundle, variations);
     return '<script src="'+url+'"></script>';
+}
+
+function entryMap(req, bundle, variations) {
+    // req.mendel.getBundleEntries contains all bundles as keys and array of
+    // entries that were used and normalized by variations, this allows apps
+    // to do their specific logic with their bundles
+    var bundles = req.mendel.getBundleEntries();
+
+    // In this particular case, we used "temporary" on the .mendelrc
+    // go expose our lazy bundle entries from extractify
+    var lazy =  [
+        '<script>',
+        '   (function(){',
+        '       var nameSpace = "_extractedModuleBundleMap";',
+        '       var url = "'+req.mendel.getURL(bundle, variations)+'";',
+        '       window[nameSpace] = window[nameSpace] || {};',
+        '       ' + bundles[bundle].map(function(entry) {
+                    return 'window[nameSpace]["'+entry+'"] = ' + ' url;';
+                }).join('\n       '),
+        '   })()',
+        '</script>'
+    ];
+    return lazy.join('\n');
 }
 
 module.exports = app;
--- a/packages/mendel-development-middleware/index.js
+++ b/packages/mendel-development-middleware/index.js
@@ -44,6 +44,40 @@ function MendelMiddleware(opts) {
     return function(req, res, next) {
         req.mendel = req.mendel || {};
 
+        req.mendel.getBundleEntries = function() {
+            console.log(bundles)
+            return Object.keys(bundles).reduce(
+
+                function(outputBundles, id) {
+                    var bundle = bundles[id];
+                    var outputEntries = [];
+
+                    [].concat(bundle.entries).filter(Boolean).forEach(
+                        function(entry) {
+                            existingVariations.forEach(function(variation) {
+                                variation.chain.forEach(function(dirPath) {
+                                    var absolutePath = path.resolve(
+                                        config.basedir, dirPath, entry
+                                    );
+                                    if (-1 === outputEntries
+                                        .indexOf(absolutePath)
+                                    ){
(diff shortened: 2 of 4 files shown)


irae:
BTW, this is the HTML the server outputs with this PR when running in production mode:
```html
<!DOCTYPE html>
<html><head></head><body>
<div id="main"></div>
<script src="/mendel/bWVuZGVsAf-eAHzTL42-AO92lVg5vzB6T2fzo0jT/vendor.js"></script>
<script src="/mendel/bWVuZGVsAQEBAf8KAL2LRWOdVDEEsryJ5lKlvdl10tNb/main.js"></script>
<script>
   (function(){
       var url = "/mendel/bWVuZGVsAQH_AgDKJJ1oIxfcES0mQD4Ao92S_QvBWw/temporary.js";
       window.lazyEntries = {
           "components/lazy.js" :  url,
       };
   })()
</script>
</body></html>
```

examples/full-example/app.js
```diff
@@ -33,7 +33,7 @@ app.get('/', function(req, res) {
             // WARNING: This temporary "lazy" bundle is not part of the example
             // The dev team needs this here for a while in order to
             // do some work in parallel branches
-            bundle(req, 'temporary', variations),
+            entryMap(req, 'temporary', variations),
```
tufandevrim: What if I have more than one seperate `lazy` bundle?

packages/mendel-middleware/index.js
```diff
@@ -24,6 +24,23 @@ function MendelMiddleware(opts) {
     return function(req, res, next) {
         req.mendel = req.mendel || {};
 
+        req.mendel.getBundleEntries = function() {
+            return Object.keys(trees.bundles).reduce(
```
tufandevrim: I think we can cache the result of this. No need to do this for every single request.

examples/full-example/app.js
```diff
@@ -47,4 +47,27 @@ function bundle(req, bundle, variations) {
...
+    // to do their specific logic with their bundles
+    var bundles = req.mendel.getBundleEntries();
+
+    // In this particular case, we used "temporary" on the .mendelrc
+    // go expose our lazy bundle entries from extractify
+    var lazy =  [
+        '<script>',
+        '   (function(){',
+        '       var nameSpace = "_extractedModuleBundleMap";',
+        '       var url = "'+req.mendel.getURL(bundle, variations)+'";',
```
tufandevrim: So there is no get away of using `findTreeForVariations` since it is used by `req.mendel.getURL`, right. At the end of the day #34 could make 3 `findTreeForVariations` calls as well.

examples/full-example/app.js
tufandevrim: For example 
`
entryMap(req, 'temporary', variations)
entryMap(req, 'temporary2', variations)
entryMap(req, 'temporary3', variations)
`
So we need to call `req.mendel.getBundleEntries()` and `findTreeForVariations` and `immediate function` 3 times right?

examples/full-example/app.js
```diff
@@ -47,4 +47,27 @@ function bundle(req, bundle, variations) {
     return '<script src="'+url+'"></script>';
 }
 
+function entryMap(req, bundle, variations) {
+    // req.mendel.getBundleEntries contains all bundles as keys and array of
+    // entries that were used and normalized by variations, this allows apps
+    // to do their specific logic with their bundles
+    var bundles = req.mendel.getBundleEntries();
```
tufandevrim: could move this line out of this function, so we dont have to call x times for x number of lazy bundles

examples/full-example/app.js
irae: Yes, every application knows about which bundle is needed.

examples/full-example/app.js
irae: As explained, yes, we need one `findTreeForVariations` per bundle that is **used** per request. But looping through all available bundles does not mean used. Also, by letting it be granular we allow people to flush the essential URLs first and eventually flush some more if they still need, making it an API that encourages performance improvements, while #34 removes this option and encourages duplicate calls to `findTreeForVariations`.

examples/full-example/app.js
irae: We can, this won't save any performance since this function executes in nanoseconds, I put it here because it reads well for people learning to have this be grouped with the next command. Also makes the request code above simpler and the params exactly the same as the `bundle` function.

packages/mendel-middleware/index.js
irae: Caching this would not help anything performance wise because this function takes nanoseconds to execute. In the other hand, if we cache it, and a consumer mutates our cache, we are screwed. Computing new arrays and new objects from scratch is safer and it is not slow at all. You can try it, do a 10k loop with this function and it will still be like 1ms in a production server.

packages/mendel-development-middleware/index.js
```diff
@@ -39,11 +39,55 @@ function MendelMiddleware(opts) {
...
+
+    console.log(allDirs);
+
     var loader = new MendelLoader(existingVariations, config, module.parent);
 
     return function(req, res, next) {
         req.mendel = req.mendel || {};
 
+        req.mendel.getBundleEntries = function() {
+            console.log(bundles)
```
tufandevrim: remove console.log

irae: I was under the impression you would continue this work and I left it there to help understanding the current state things are working.

irae: ok, ill remove and ad tests to this PR

PR #36, opened 2016-08-20
Groundwork to support extractify-like algorithm on Mendel CLI; Flattens arrays, verbose mode and normalize entries

irae:
This fixes the verbose output by default, making it an option.
It also makes it possible to do nested arrays in .mendelrc files, which is specifically useful for yaml files. Here is a simple example:

```yml
sharedConfig:
  vendorLibs: &vendorLibs
    - react
    - react-dom
    - react-nice-plugin
    - lodash/is/cool
  onDemandBundle: &onDemandBundle
    - ./src/less-used-feature.jsx

bundles:
  main:
    entries:
      - ./src/main.js
    external:
      - *vendorLibs
      - *onDemandBundle
  less-used:
    external: *vendorLibs
    require: *onDemandBundle
```

We are also deprecating use of relative paths on entries with a warn message.

@tufandevrim can you review?

### DIFF
--- a/bin/index.js
+++ b/bin/index.js
@@ -15,20 +15,31 @@ var applyExtraOptions = require('mendel-development/apply-extra-options');
 var mendelBrowserify = require('mendel-browserify');
 var mendelRequirify = require('mendel-requirify');
 
+var debug = process.argv.some(function(arg) {
+    return /--verbose/.test(arg);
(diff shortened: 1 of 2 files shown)


tufandevrim:
:+1: LGTM

PR #37, opened 2016-08-21
Post manifest architecture, mendel-manifest-extract-bundles and example on how to use it

irae:
This PR is a bit large but it is easy to understand commit by commit.

The first one is pretty straight forward, it just moves code around to be used on the later commits.

The second commit introduces a new `.mendelrc` feature, the `manifestProcessors`. Here is how it is supposed to be used:

```yaml
manifestProcessors:
  - ./my-path/post-process-manifest.js
  -
    - postprocessor-npm-module
    - option1: strvalue
      option2: true
```

Any post processor will be loaded relative to `basedir` or from the project `package.json`. And every post processor has the following signature:

```js
module.exports = function processorName(allManifests, options, next) {
    // do your stuff with all manifests
   next(newManifestObject);
};
```

All manifests are also re-sorted and validated by the same functions we use in other places in the codebase (extracted in the first commit.

The last commit introduces a new package `mendel-manifest-extract-bundles` that does the same as extractify, but only as a post processor for mendel.

Here is a simplified version of the `full-example` code that uses this new processor: 

```yaml
bundles:
  main:
    entries:
      - ./isomorphic/base/main.js
    external:
      - ./isomorphic/base/components/lazy.js
  lazy:
    require:
      - ./isomorphic/base/components/lazy.js
    ## not needed anymore
    ## external:
    ##   - ./isomorphic/base/components/button.js

manifestProcessors:
  -
    - mendel-manifest-extract-bundles
    - external: lazy
      from: main
```

The externals are simple in our full example, but can be hundreds of files in real applications.

You can also run `mendel --verbose` to see a lot of useful information to understand the code better.

@tufandevrim

### DIFF
--- a/bin/index.js
+++ b/bin/index.js
@@ -15,6 +15,8 @@ var applyExtraOptions = require('mendel-development/apply-extra-options');
 var mendelBrowserify = require('mendel-browserify');
 var mendelRequirify = require('mendel-requirify');
 
+var postProcessManifests = require('./post-process-manifest');
+
 var debug = process.argv.some(function(arg) {
     return /--verbose/.test(arg);
 });
@@ -34,6 +36,7 @@ var config = ['basedir','outdir','bundlesoutdir','serveroutdir',
     });
 
 if(debug) {
+    rawConfig.verbose = true;
     config.verbose = true;
     console.log('mendel parsed config');
     logObj(config);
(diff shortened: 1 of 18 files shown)


tufandevrim:
👍

PR #38, opened 2016-09-02
Updated ReadMe for readability

jamesalley:
I tried not to change the intent at all, but to update for readability, parallel structures, verb tense, etc. Nice work. The opening description and the Mendel paragraph were beautiful!


irae:
Thanks a lot!

PR #40, opened 2016-10-13
req.mendel.resolver requires bundles array to improve performance

tufandevrim:
This PR intends to improve performance of `loader.resolver` function by only checking bundle list in `findServerVariationMap` function of `mendel-core/trees`

`findServerVariationMap` does not have to go thru every single bundle id specified in `.mendel.rc`

@irae

### DIFF
--- a/packages/mendel-core/trees.js
+++ b/packages/mendel-core/trees.js
@@ -42,7 +42,7 @@ MendelTrees.prototype.findTreeForVariations = function(bundle, variations) {
     return finder.found();
 };
 
-MendelTrees.prototype.findServerVariationMap = function(variations) {
+MendelTrees.prototype.findServerVariationMap = function(bundles, variations) {
     if (typeof variations ==='string') {
         variations = [variations];
     }
@@ -53,7 +53,11 @@ MendelTrees.prototype.findServerVariationMap = function(variations) {
     var lookupChains = self._buildLookupChains(variations);
     var finder = new MendelServerVariationWalker(lookupChains, base);
 
-    Object.keys(self.bundles).forEach(function (bundle) {
+    function selectBundles(key) {
+        return bundles.indexOf(key) !== -1;
+    }
+
+    Object.keys(self.bundles).filter(selectBundles).forEach(function (bundle) {
         self._walkTree(bundle, finder);
         variationMap = xtend(variationMap, finder.found());
     });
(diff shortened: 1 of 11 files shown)


packages/mendel-core/trees.js
```diff
@@ -42,7 +42,7 @@ MendelTrees.prototype.findTreeForVariations = function(bundle, variations) {
     return finder.found();
 };
 
-MendelTrees.prototype.findServerVariationMap = function(variations) {
+MendelTrees.prototype.findServerVariationMap = function(variations, bundleList) {
```
irae: If you take a look at the other methods of this class, many methods receive a singe bundleId as first param and variations as second param. Can we change the param order?

```js
MendelTrees.prototype.findServerVariationMap = function(bundle, variations) {
```

Do it need it to be a list? I don't know if we really need it to be a list. If we need, it's ok, otherwise we could just pass a single ID and the API would be very similar to the other functions, right?

irae:
Also, can you please bump the version on package.json for mendel core and mendel loader?

packages/mendel-core/trees.js
tufandevrim: > If you take a look at the other methods of this class, many methods receive a singe bundleId as first param and variations as second param. Can we change the param order?

Sounds good, I will change the param order

> Do it need it to be a list? I don't know if we really need it to be a list. If we need, it's ok, otherwise we could just pass a single ID and the API would be very similar to the other functions, right?

Yes we needed. Otherwise it will walk thru every single bundle. This is the major perf improvements for very big projects with multiple bundles on the ssr rendering

tufandevrim:
> Also, can you please bump the version on package.json for mendel core and mendel loader?

Sure will do

packages/mendel-core/trees.js
irae: Yes, I know a new param is needed.

My question was about the param type. Instead of "bundles" array, could it be "bundle" string.

In other words: can the new param be `findServerVariationMap("my_app", ["base"])` or does it need to be `findServerVariationMap(["my_app", "my_other_app"], ["base"])` ?

irae:
Thanks very much for this patch, hope we can soon have the next step in perf improvements. Just waiting for the version bump and mandatory array so we can merge and publish.

PR #41, opened 2016-10-13
Merge pull request #1 from yahoo/master

shalomvolchok:
Merging updates from yahoo


shalomvolchok:
Sorry, meant to pull into my fork. Great library!!

irae:
Hey! Not a problem! We are heavily working on Mendel 2.0 with a lot of great stuff incoming. I am publishing a lot of new documentation and will post the new architecture design soon.

It is a pleasure to know we have people interested in Mendel!

irae:
Don't worry, most stuff will be retro-compatible, so it is safe to use the current version right now if you'd like.

shalomvolchok:
Yeah for sure! We have started playing with Mendel and I don't know of anything else like it. The flat folder/file structure, the server side require resolution, the dynamic bundle resolution and creation, variation inheritance, the total decoupling from assignment & tracking, and I know it does a ton more... really amazing work!! Thanks for open sourcing this!!

One question, if you don't mind. I'd like to know if I'm starting to understand layers correctly... Am I correct to think that a layer is a collection of mutually exclusive variations; where variations are defined, in a `.mendelrc` file, as one or more folders from the `variationsdir` (using variation inheritance, if desired)?

Let's say we have a file structure like this:

```
bash> tree

├── base
│   ├── file1.js
│   └── file2.js
└── variations
    ├── bucket_A
    │   └── file1.js
    ├── bucket_B
    │   └── file1.js
    ├── bucket_C
    │   └── file2.js
    └── bucket_D
        └── file2.js

```

And in our `.mendelrc`:
```
# ... #
variationsdir: variations
variations:
    bucket_A:
    bucket_B:
    bucket_C:
    bucket_D:
```

And now we want to do the assignment part... Would the following, conceptual example, make sense? 

```js

// false as a placeholder for base
const layer1 = [false, "bucket_A", "bucket_B"];
const layer2 = [false, "bucket_C", "bucket_D"];

const userId = 1234;

const variations = [
    hashUserIdToSingleElementInArray(userId, layer1),
    hashUserIdToSingleElementInArray(userId, layer2)
].filter(selection => !!selection)

var resolver = req.mendel.resolver(variations);

/* ... */
```

Does this seem like the right direction to be thinking in?

irae:
Yeah, that seems to make sense. I am not sure using `userId` will give you the correct uniform distribution you want, specially if `userId` is an incremental integer, this could mean new users get on an experiment consistently and old users consistently on base, and that's not a neutral distribution. But assuming the function `hashUserIdToSingleElementInArray` is a fair random assignment implementation, balanced the way you would like, your code should work fine.

Also — I think you already realized based on your example — once you go multilayer you can have conflicts, and if you poke around the code enough you will find that `mendel-core` already detects that. For instance, If in you next revision `bucket_C` starts implementing a variation of `file1.js`, it will conflict with `bucket_A` or `bucket_B`. MendelMiddleware currently ignores the conflicts, but you can get a list of runtime conflicts if you use `findTreeForVariations` directly.

PS: `.filter(selection => !!selection)` might read better as `.filter(Boolean)` 😉

shalomvolchok:
Great, thanks for the confirmation and elaboration about conflicts.  I had noticed that with conflicts Mendel seemed to choose the file based on order in the `.mendelrc` file. I guess this is what you mean by MendelMiddleware ignores conflicts. So it seems the issues with getting conflicts is not that Mendel will break, but that an experiment could inadvertently not be the combination we intend. I played around with `findTreeForVariations`. That seems very useful. Thanks for pointing it out! Something like this, yeah?
```js
require('mendel-core/trees')()
    .findTreeForVariations('main', ['bucket_A', 'bucket_C'])
    .conflictList
```

Yes, in production we'd do our best to make sure it's a fair random assignment and also that we don't get correlations between layers by only hashing on the userId. Probably in this case we'd hash on a string of userId `uuid.v4()` and the layer name ``${userId}.${layerName}``.  We have been working with the Planout library from facebook. So the real thing would probably end up looking more like this, assuming we wanted a uniform assignment:

```js
const userId = uuid.v4(); 
const layerName = "layer1";
const layer = [false, "bucket_A", "bucket_B"];
const hash = sha1(`${userId}.${layerName}`);

const randIndex = new BigNumber(hash.substr(0, 15), 16).modulo(layer.length).toNumber();
const assignment = layer[randIndex];
```
That seem better?

PS: `.filter(Boolean)` nice one :+1:

irae:
LGTM

I guess you can use crypto directly and get this sha1 as buffer and slice the binary without turning it to string before converting to number, but I digress with over-optimizations =)

Here at Yahoo we use our proprietary assignment backend and I don't know much of the internals. In theory Mendel should work very well with Planout, but I didn't have time to try it out. This would be an interesting example to have on the repo and in the docs.

shalomvolchok:
:+1: for over-optimization

would you like a PR that adds a Planout example?

irae:
I would be very please indeed! Any small contribution is very welcome, and an example with Planout is specially welcome, since we don't have any example with random assignment solutions out there!

PR #43, opened 2016-10-18
Adding cache to getUrl mendel operation

tufandevrim:
This PR intends to improve the performance of mendel middleware getUrl operation by utilizing an object cache. If memory leak is the concern, we could add a limit to cache size. Eventually we could also add `useCache` flag in `.mendel.rc` to make it optional.
@irae

### DIFF
--- a/packages/mendel-middleware/index.js
+++ b/packages/mendel-middleware/index.js
@@ -24,6 +24,8 @@ function MendelMiddleware(opts) {
         parentModule: module.parent
     });
 
+    var urlCache = {};
+
     return function(req, res, next) {
         req.mendel = req.mendel || {};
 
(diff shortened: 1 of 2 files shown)


packages/mendel-middleware/index.js
```diff
@@ -45,8 +47,18 @@ function MendelMiddleware(opts) {
         };
 
         req.mendel.getURL = function(bundle, variations) {
-            var tree = trees.findTreeForVariations(bundle, variations);
-            return getPath({ bundle: bundle, hash: tree.hash });
+            var cacheKey = bundle + '+' + variations.sort().join('+');
```
irae: Supposing we are doing caching, which I am not in favor of, there is a better option for cache key than just sorting the string.

We should us the result for `_findChainForVariations` or something very similar. The resolved variations are a better way than the input variations for a number of reasons:

  1. The resolved variations are consistent and sorted. For instance, if the application passes `['foo', 'bar', 'baz']` and 'bar' is not a valid variation, the cache key would become `baz+foo`.
  2. The chain resolved on `mendel-core` is consistent order with the `.mendelrc` which makes it easy for debug (the order is what the code will be resolved to, as opposed to alphabetical order).
  3. Also, it makes all of the following resolve to same key: `['foo', 'bar', 'baz']`, `['foo', 'baz']`, `['baz', 'foo', 'baz']`, `['baz', 'foo']`, etc. This would reduce memory footprint and would reduce CPU use (for variations that are effectively the same, although input is different.

But to leverage this consistent hash we need to change how we interact with `mendel-core`, not just on the middleware.

PR #44, opened 2016-10-18
Remove unused devdep from pkg json

tufandevrim:
`npm install` returns 404  'mendel-extractify' is not in the npm registry. As far as I know it is not used anymore. We could use this PR to remove it completely from the repo, as well.


irae:
Yes, we can remove it, but also if you run `npm link all` it will be linked and won't hit the 404. I'll remove it for now. Thanks

PR #46, opened 2016-10-22
Add better explanation for non-deterministic file errors

irae:
This changes the error message on `mendel-browserify` to be more friendly and link to in-depth documentation for this very hard to debug error.


yahoocla:
CLA is valid!

PR #47, opened 2016-10-22
A number performance improvements on mendel-middleware and mendel-core

irae:
* Mendel Core updated ~~so server resolver don't hash tree for no reason~~ reducing memory allocation and "pushing" sha1 -- sha was never actually computed in the first place, good job @gomezd -- What I did removes just a tiny bit, what David did was the bulk of it already.
  * Mendel Core new public method variationsAndChains
  * Mendel Middleware new public method.setVariations()
  * Mendel Middleware new public method .getBundle()
  * Mendel Middleware .getURL() changed so variations is deprecated

With those changes, Mendel Core is now out of the box more performant, maybe the last chages with resolver receiving a bundles array might be even obsolete. Benchmark required to confirm.

Mendel Middleware new methods are intended for better caching by two improvements:

  1. By requiring per-request `setVariations` we can now assure transparently that each bundle only computes the hash once.
  2. By exposing `req.mendel.setVariations()` or directly through `req.mendel.variations` apps have a way of generating consistent and efficient hash that won't duplicate or cache extraneous keys based on "input variations", instead relying on "resolved variations".
  3. New method `req.mendel.getBundle()` allow apps to create powerfull introspection APIs or even better custom debugging tools, this method will also be used by Mendel 2.0 resolver to avoid extra traversal of bundles, so we are commiting to this API upfront.

Missing for future commits before merging this changes:

 - [x] more tests
 - [x] mendel-development-middleware compatible API changes
 - [x] bump package.json in all affected modules

After some internal tests, Julien Lecomte pointed out that the first version of this PR without caching was slightly worst than master.

So first, I added a small bash script on master (to have a baseline) and started tweaking. Anyone can run the script against the full example:

    $ ./bench.sh

I also added a param to full example to avoid caching if desired:

    $ NODE_ENV=production CACHE=false node .

I did some investigation and here are the two optimizations I did:

  1. I found out that `Object.freeze`, even on a per process based was slowing down the implementation too much. This means traversing a frozen Object is more expensive than traversing a regular object.
  2. Some extra tweaks to call some sub-routines less often and use primitive params instead of objects.

Results (each test is 1k iterations, I run each branch 5 times and picked the best result of each):

```
Percentage of the requests served within a certain time (ms)
master        | PR-no-cache   | PR-with-chache
              |               |
  50%      7  |   50%      7  |   50%      2
  66%      8  |   66%      7  |   66%      2
  75%      8  |   75%      8  |   75%      2
  80%      8  |   80%      8  |   80%      2
  90%      8  |   90%      8  |   90%      2
  95%      9  |   95%      9  |   95%      2
  98%     11  |   98%     11  |   98%      3
  99%     13  |   99%     13  |   99%      7
 100%     14  |  100%     15  |  100%      8
              |               |
 mean  7.773  |  mean  7.575  |  mean  2.044 
```

In general this PR is an improvement, but for some reason about 0.1% of requests are 1ms slower. I suspect this is when server is under super heavy load and GC kicks in. I won't investigate further since start entering the realm of over-optimization.

### DIFF
--- a/examples/full-example/app.js
+++ b/examples/full-example/app.js
@@ -6,9 +6,17 @@ var ReactDOMServer = require('react-dom/server');
 var express = require('express');
 var logger = require('morgan');
 var MendelMiddleware = require('mendel-middleware');
+var cache = true;
+
 if (process.env.NODE_ENV !== 'production') {
     MendelMiddleware = require('mendel-development-middleware');
+    cache = false;
 }
+
+if (String(process.env.CACHE) === 'false') {
+    cache = false;
+}
+
 
 var app = express();
 app.use(logger('tiny'));
@@ -20,6 +28,9 @@ app.get('/', function(req, res) {
     .split(',').filter(Boolean);
     var serverRender = req.query.ssr !== 'false' && req.mendel.isSsrReady();
     var optionalMarkup = "";
+
+    // populates req.mendel.variations with validated and sorted variations
+    req.mendel.setVariations(variations);
 
     if (serverRender) {
         // To improve ssr performance, you need to pass
@@ -34,12 +45,12 @@ app.get('/', function(req, res) {
         '<!DOCTYPE html>',
         '<html><head></head><body>',
             '<div id="main">'+optionalMarkup+'</div>',
-            bundle(req, 'vendor', variations),
-            bundle(req, 'main', variations),
+            bundle(req, 'vendor'),
+            bundle(req, 'main'),
             // The full example supports on-demand loading, lazy bundle
             // is only loaded client-side when a button is clicked in the
             // application
-            entryMap(req, 'lazy', variations),
+            entryMap(req, 'lazy'),
         '</body></html>'
     ].join('\n');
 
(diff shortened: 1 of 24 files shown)


yahoocla:
CLA is valid!

irae:
This confirms my hypothesis of garbage collector being the culprit of the 100% percentile.

This is unlikely to happen in production, since I would expect the garbage collector to trigger in idle times, not in execution time.

screenshot 2016-10-25 at 11 14 34 am

irae:
Forgot to mention, above graph is with NODE_ENV=production and cache is used on the full example.

packages/mendel-middleware/index.js
```diff
@@ -44,12 +47,54 @@ function MendelMiddleware(opts) {
...
+                var varsAndChains = trees.variationsAndChains(variations);
+                req.mendel.variations = varsAndChains.matchingVariations;
+                req.mendel.lookupChains = varsAndChains.lookupChains;
+            }
+            return req.mendel.variations;
+        };
+
+        req.mendel.getBundle = function(bundle) {
+            if (!req.mendel.variations) {
+                throw new Error('Please, call req.mendel.setVariations first');
```
stephanwlee: `Please, call` -> `Please call`

examples/full-example/app.js
```diff
@@ -6,10 +6,18 @@ var ReactDOMServer = require('react-dom/server');
...
 var logger = require('morgan');
 var MendelMiddleware = require('mendel-middleware');
+var cache = true;
+
 if (process.env.NODE_ENV !== 'production') {
     MendelMiddleware = require('mendel-development-middleware');
+    cache = false;
+}
+
+if (String(process.env.CACHE) === 'false') {
```
stephanwlee: How about a little prefix like `MENDEL_CACHE` like Babel? http://babeljs.io/docs/usage/require/#babeldisablecache

examples/full-example/app.js
```diff
@@ -34,46 +45,66 @@ app.get('/', function(req, res) {
...
-        '   })()',
-        '</script>'
-    ];
-    return entryMapScript.join('\n');
+
+function entryMap(req, bundle) {
+    var key = bundle + ':' + req.mendel.variations.join(':');
+    if (!cache || !entryMapCache[key]) {
+
+        // req.mendel.getBundleEntries contains all bundles as keys and array of
```
stephanwlee: Awkward sentence. 

From
```
req.mendel.getBundleEntries contains all bundles as keys and array of
entries that were used and normalized by variations, this allows apps
to create specific logic with their bundles in runtime
```

To
```
`req.mendel.getBundleEntries` contains all bundles as keys and arrays of
entries that were used (normalized by variations) as values. This allows apps
to create a specific logic with their bundles in the runtime.
```

I might be wrong about your intention.

packages/mendel-core/tree-variation-walker.js
```diff
@@ -11,9 +11,9 @@ var MendelWalker = require('./tree-walker');
 util.inherits(MendelVariationWalker, MendelWalker);
 module.exports = MendelVariationWalker;
 
-function MendelVariationWalker(_lookupChains, _base) {
+function MendelVariationWalker(_lookupChains, _base, _hash) {
```
stephanwlee: The `_hash` is not used anywhere. Am I correct?

packages/mendel-core/trees.js
```diff
@@ -20,37 +20,27 @@ function MendelTrees(opts) {
...
 }
 
-MendelTrees.prototype.findTreeForVariations = function(bundle, variations) {
-    if (typeof variations ==='string') {
-        variations = [variations];
-    }
-
-    var lookupChains = this._buildLookupChains(variations);
+MendelTrees.prototype.findTreeForVariations = function(bundle, lookupChains) {
     var finder = new MendelVariationWalker(lookupChains, this.config.base);
```
stephanwlee: About the comment above: hash is not passed here so I am confused :)

packages/mendel-development-middleware/index.js
```diff
@@ -86,12 +88,38 @@ function MendelMiddleware(opts) {
...
+                req.mendel.variations = variations;
+            }
+            return req.mendel.variations;
+        };
+
         req.mendel.getURL = function(bundle, variations) {
-            var vars = variations.join(',') || config.base;
+            if (!req.mendel.variations && variations) {
+                console.warn(
+                    'mendel.getURL(bundle, variations) is deprecated. '+
```
stephanwlee: uglier warning message like `[DEPRECATED]` seems like an industry norm.

packages/mendel-core/tree-variation-walker.js
irae: it is used in super, this is pass through here, if we don't collect and someone calls without `new` it would break. Welcome to ES5 quirks, this is a detail of how to do what ES6 classes abstract for you.

packages/mendel-core/trees.js
irae: In this particular instance we are ok with the default (hash is true). The constructor needs to support it. See the tree-walker to get it.

examples/full-example/app.js
irae: It is a very contrived example, not an API supported by Mendel... but I'll change it anyway.

PR #50, opened 2016-11-07
Linkall without losing color

stephanwlee:
spawn allows you to mock the io like ansi so `npm` doesn't print in no-color.

### DIFF
--- a/scripts/linkall.js
+++ b/scripts/linkall.js
@@ -5,13 +5,13 @@
 
 var fs = require('fs');
 var path = require('path');
-var exec = require('child_process').exec;
+var spawn = require('child_process').spawn;
 
 var packagesdir = path.join(__dirname, '../packages');
 var examplesdir = path.join(__dirname, '../examples');
 var rootdir = path.join(__dirname, '..');
 
-npm('install async', rootdir, function() {
+npm(['install', 'async'], rootdir, function() {
     var async = require('async');
     var linkedModules = [];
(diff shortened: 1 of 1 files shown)


yahoocla:

yahoocla:
CLA is valid!

package.json
```diff
@@ -33,6 +33,7 @@
     "unit": "tap test/*.js",
     "unit-file": "tap",
     "linkall": "node ./scripts/linkall.js",
+    "unlinkall": "find . -name a -type d -exec rm -rf {} \\;",
```
irae: ???

stephanwlee: :\ I felt like there should be the opposite of linkall

packages/mendel-cli/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "mendel",
   "version": "1.3.0",
   "description": "Mendel command line utility",
+  "preferGlobal": true,
```
irae: Not sure about this.

package.json
irae: Yes, but I think we need to do it differently. I think we need to list all linked mendel-* modules and remove them. For instance, when switching from v2 to master we should be able to unlink the v2 modules without having them on the tree, right?

irae: Wait... what is "a"? confused, maybe you know better how npm works with links 😜

PR #51, opened 2016-11-08
Short ID and exposes for privacy and compression improvements

irae:
In some cases, full path was exposed on module names, specially if you use mendel-manifest-extract-bundles on a bundle with node_modules inside of it.

This fixes this by changing both ids, expose and deps to remove:
1. Basedir
2. Anything before first "node_modules" string on path

In most cases both are the same, but since plugins can include shims on their source code (like browserify shims) checking for the first occurrence of node_modules improves the path normalization.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "dependencies": {},
   "devDependencies": {
     "async": "^2.1.2",
+    "browserify": "^13.1.1",
     "eslint": "^2.3.0",
     "eslint-cli": "^1.1.0",
     "eslint-plugin-react": "^5.2.2",
--- a/packages/mendel-cli/mendel-browserify-runner.js
+++ b/packages/mendel-cli/mendel-browserify-runner.js
@@ -49,13 +49,11 @@ module.exports = function(rawConfig, options) {
     }
 
     if (Array.isArray(options.variations)) {
-        console.log(options.variations);
         var old = mendelConfig.variations;
         mendelConfig.variations = {};
         options.variations.forEach(function(id) {
             mendelConfig.variations[id] = old[id];
         });
-        console.log(mendelConfig.variations);
     }
 
     bundles.forEach(function(bundle) {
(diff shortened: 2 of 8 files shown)


yahoocla:
CLA is valid!

package.json
```diff
@@ -48,6 +48,7 @@
   "dependencies": {},
   "devDependencies": {
     "async": "^2.1.2",
+    "browserify": "^13.1.1",
```
irae: This was a missing peer-dependency of mendel-cli. Breaks tests if you checkout a clean clone and run tests.

packages/mendel-cli/mendel-browserify-runner.js
```diff
@@ -49,13 +49,11 @@ module.exports = function(rawConfig, options) {
     }
 
     if (Array.isArray(options.variations)) {
-        console.log(options.variations);
         var old = mendelConfig.variations;
         mendelConfig.variations = {};
         options.variations.forEach(function(id) {
             mendelConfig.variations[id] = old[id];
         });
-        console.log(mendelConfig.variations);
```
irae: oops, cleaning

packages/mendel-cli/mendel-browserify-runner.js
```diff
@@ -49,13 +49,11 @@ module.exports = function(rawConfig, options) {
     }
 
     if (Array.isArray(options.variations)) {
-        console.log(options.variations);
```
irae: oops, cleaning

scripts/linkall.js
```diff
@@ -11,7 +11,7 @@ var packagesdir = path.join(__dirname, '../packages');
 var examplesdir = path.join(__dirname, '../examples');
 var rootdir = path.join(__dirname, '..');
 
-npm(['install', 'async'], rootdir, function() {
+npm(['install'], rootdir, function() {
```
irae: This also helps run test with a clean close of the repo. Async is now on devDependencies, so no need to force install separately.

test/mendel-browserify.js
```diff
@@ -5,7 +5,7 @@
 var path = require('path');
 var mkdirp = require('mkdirp');
 var test = require('tap').test;
-var mendelPlugin = require('mendel-browserify');
+var mendelPlugin = require('../packages/mendel-browserify');
```
irae: Just in case we run tests without running linkall.

stephanwlee:
👍 Good!

PR #52, opened 2016-11-10
Refactor on the Watch+FS and analytics

stephanwlee:
1. Now tree=watch+fs is separated into separate steps.
2. The reason why pretty printer was broken is... text-table re-aligns based on max number of characters on each rows and it was counting bash coloring part of the string as length (something that goes like \[30;) and was aligning so badly. Took it out.

New screenshot looks like following:
screen shot 2016-11-09 at 11 41 39 pm

### DIFF
--- a/packages/mendel-pipeline/src/helpers/analytics/cli-printer.js
+++ b/packages/mendel-pipeline/src/helpers/analytics/cli-printer.js
@@ -1,14 +1,7 @@
 const BasePrinter = require('./printer');
 const chalk = require('chalk');
-const table = require('text-table');
 const prettyMs = require('pretty-ms');
 const figure = require('figures');
-
-function println(text, indentation) {
-    const indent = new Array(indentation + 1).join('  ');
-
-    console.log(indent + text.split('\n').join('\n' + indent));
-}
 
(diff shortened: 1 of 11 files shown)


yahoocla:

yahoocla:
CLA is valid!

packages/mendel-pipeline/src/helpers/analytics/cli-printer.js
```diff
@@ -54,48 +47,59 @@ class CliPrinter extends BasePrinter {
...
             const dataPart = subData[groupedName];
+
+            // Special case: make the labels pretty!
+            if (indentation === 0 && dimensions.length === 1 && dimension !== 1) {
+                // Example: <pid> (groupNameA+groupNameB+groupNameC+etc…)
+                const set = new Set();
+                dataPart.forEach(p => set.add(p.name.split(':')[1]));
+                const groupNames = Array.from(set.keys()).slice(0, 3).join('+') + (set.size > 3 ? '+etc…' : '');
+                groupedName += ` (${groupNames})`;
+            }
```
irae: Lol, it does not look that much difference, we were doing most calculation ourselves anyway.

PR #53, opened 2016-11-12
Proposal of per enviroment pipeline without master/slave process

irae:
@stephanwlee here is my idea to handle the problem we discussed today

Everything that was small fixes or small improvements that could be merged on master I did. Additional comments inline.

### DIFF
--- a/packages/mendel-pipeline/src/cli.js
+++ b/packages/mendel-pipeline/src/cli.js
@@ -1,8 +1,8 @@
 #!/usr/bin/env node
-const mendelPipeline = require('./pipeline');
-const config = require('../../mendel-config');
+/* eslint max-len: "off" */
 const program = require('commander');
 const path = require('path');
+const MendelPipelineDaemon = require('./daemon');
 
 process.env.MENDELRC = process.env.MENDELRC || '.mendelrc_v2';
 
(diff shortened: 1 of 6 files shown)


yahoocla:
CLA is valid!

packages/mendel-pipeline/src/cli.js
```diff
@@ -20,6 +20,30 @@ program
...
-// Example usage
-const mendelConfig = config(Object.assign(program, {cwd: path.resolve(process.cwd(), program.args[0])}));
-mendelPipeline(mendelConfig);
+
+const cliOptions = Object.assign(program, {
+    cwd: path.resolve(process.cwd(), program.args[0]),
+});
+
+if (program.watch) {
+    const daemon = new MendelPipelineDaemon(cliOptions);
```
irae: See the `startServer` idea in the `daemon.js`, since there will be a default port, this would throw "ERRADDINUSE" and the program exists, no harm done.

packages/mendel-pipeline/src/pipeline.js
```diff
@@ -1,61 +1,96 @@
...
+                    const total = this.state.registry.size();
+                    this._debug(`${doneEntries} entries were processed.`);
+                    this._debug(`${total} entries in registry.`);
+
+                    startedEntries = 0;
+                    doneEntries = 0;
+
+                    this.emit('idle', doneEntries);
+                }
+            });
```
irae: I think this is a healthy and stable lifecycle to debug on the daemon. I think we can keep this even if we decide later the daemon is not an emitter, or that the `idle` event should not exist.

packages/mendel-pipeline/src/cli.js
```diff
@@ -20,6 +20,30 @@ program
...
+            if (error && isConnectionRefused(error)) {
+                const daemon = new MendelPipelineDaemon(cliOptions);
+                daemon.run();
+            } else {
+                request({
+                    url: defaultDeamonAddres + '/run',
+                    query: cliOptions,
+                });
+            }
+        });
```
irae: Both the Middleware and the UnitTest can use the same API, some endpoints would be for environments + bundles (and understand the bundle config) other APIs would be per environment + entryId and only need to understand the registry.

packages/mendel-pipeline/src/daemon.js
```diff
@@ -0,0 +1,83 @@
...
+        pipeline.run(environment);
+    }
+
+    startServer(pipeline) {
+        debug(`TODO: not implemented startServer(pipeline) ${pipeline}`);
+    }
+
+    bindOutlets(pipeline) {
+        debug(`TODO: not implemented bindOutlets(pipeline) ${pipeline}`);
+    }
```
irae: Both methods above would use async methods such as `registry.getTree({trasnformIds, entryId})` or `registry.getTree({environment ,entryId})` or even more generic such as `registry.getTree(entryId, selectorFn)`.

irae:
We can still use the branch, just closing the PR for now, as per discussion that this is not really the implementation we are going towards, but the multi-pipeline might still happen without the deamon.

PR #54, opened 2016-11-15
More fully featured manifest-extract-bundles

irae:
This makes sure we remove files consistently regardless of variations, it cleans up orphans and prevent problems where some variations are lazy and some variations are not.


yahoocla:
CLA is valid!

stephanwlee:
👍

PR #55, opened 2016-11-15
Demoted deps as tools based on modified arch

stephanwlee:


yahoocla:

yahoocla:
CLA is valid!

packages/mendel-pipeline/src/pipeline.js
```diff
@@ -19,15 +19,16 @@ function MendelPipeline(options) {
...
     const transformer = new Transformer(options.transforms, options);
+    const depsResolver = new DepResolver(options);
 
-    // Pipeline steps
-    const initializer = new Initialize({registry, transformer}, options);
-    const watcher = new Watcher({registry, transformer}, options);
-    const reader = new Reader({registry, transformer}, options);
-    const ist = new IST({registry, transformer}, options);
-    const depsResolver = new DepResolver({registry, transformer}, options);
+    const toolset = {registry, transformer, depsResolver};
```
irae: You really did it? if deps will be an accessory, why don't deps run inside the IST/GST workers instead? we would save the AST part, right?

Just asking out of curiosity, really.

irae:
👍

packages/mendel-pipeline/src/pipeline.js
stephanwlee: If we impose a restriction that every plugin has to operate on AST (potentially one type?) and output AST, then yes. If not, no.

PR #56, opened 2016-11-17
Implement simple case of `normalizedId`. Works with most reasonable projects

irae:
This PR is quite big and has a lot of chance for conflicts, so we should try to merge pretty fast.

@stephanwlee I suggest you read the commits in order, the last commit is already big in itself, but the first two are way simpler. Also, the 3 commits have large contextual commit messages.

Here is a summary of changes:

1. Make `projectRoot` consistent across all files and remove the option to pass in a folder to scrape. You can still pass in a directory, but we will look for mendelrc there, not source files directly.
2. Move all variation configuration logic and parsing to `mendel-config`. Also make most configs POJO so we don't have functions to serialize between processes as a nested object.
3. Implement `normalizedId` itself

Next step after this is merged is to code the colision detection when `package.json` is used inconsistently between variations.

### DIFF
--- a/packages/mendel-pipeline/src/cache/index.js
+++ b/packages/mendel-pipeline/src/cache/index.js
@@ -1,29 +1,29 @@
 const path = require('path');
 const Entry = require('./entry.js');
+const variationMatches = require('mendel-development/variation-matches');
 
 class MendelCache {
     constructor(config) {
         this._store = new Map();
         this._baseConfig = config.baseConfig;
-        this._variationConfig = config.variationConfig;
-
-        const variationDirSet = new Set();
-        Object.keys(this._variationConfig.variations)
-            .filter(varKey => this._variationConfig.variations[varKey])
-            .forEach(varKey => {
-                this._variationConfig.variations[varKey].forEach(varFolderName => variationDirSet.add(varFolderName));
-            });
-        const varDirNames = Array.from(variationDirSet.keys());
-
-        this.variationalRegex = new RegExp(`(${varDirNames.map(dirName => `(${dirName})${path.sep}\\w+`).join('|')}|${this._baseConfig.dir})${path.sep}?`);
+        this._variations = config.variationConfig.variations;
     }
 
     getNormalizedId(id) {
-        if (isNodeModule(id)) return id;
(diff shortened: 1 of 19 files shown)


yahoocla:
CLA is valid!

packages/mendel-pipeline/src/cache/index.js
```diff
@@ -1,29 +1,29 @@
...
 
     getNormalizedId(id) {
-        if (isNodeModule(id)) return id;
-
-        // This is wrong WIP
-        // return id.replace(this.variationalRegex, '');
-        return id;
+        let normalizedId = id;
+
+        const match = variationMatches(this._variations, id);
```
stephanwlee: For my understanding of the scope: does this commit address the case when package.json include path to a file that is not in the same dir?

packages/mendel-resolver/variational-resolver.js
```diff
@@ -1,32 +1,51 @@
...
         super({cwd: projectRoot, basedir, envNames});
 
-        // Must be a path relative to the basedir
+        this.variationChain = variations.map(variationPath => {
+            return path.resolve(projectRoot, variationPath);
+        });
+        this.baseVarDir = path.resolve(projectRoot, baseConfig.dir);
+
+        this.variation = variation;
+        this.normalizedId = normalizedId;
```
stephanwlee: disregarding the fact that this property is not used, how will this be used for this package? resolving, in my view, has nothing to do with normalizedId. Intentional?

irae: I had an alternative implementation with `normalizedId` until I realized it does not always means the same thing. I was trying to get some stuff for free, without calling the resolver. For now we can remove, I guess if we do those optimizations, it will be on a higher level.

packages/mendel-pipeline/src/cache/index.js
irae: Not yet. But I don't think it will be part of this method to validate that, that method may be final and all the validation when the entry is being created. Will get to see when implementing such validations. The current implementation only covers the large portion of use cases, but anything that needs the value inside the `package.json` is incorrect so far.

PR #57, opened 2016-11-17
Only cash if result is found

juandopazo:
@jlecomte


yahoocla:
CLA is valid!

jlecomte:
👍 SHIP IT!

PR #63, opened 2016-11-29
Working draft of GST2

stephanwlee:
@irae Please advise on logical wrongdoing. I am not sure if the GST code is correct :\


yahoocla:
CLA is valid!

packages/mendel-pipeline/src/step/gst.js
```diff
@@ -0,0 +1,174 @@
...
+
+            // We need to create proxy for limiting API surface for plugin writers.
+            const context = this.getContext(nextTransformIds);
+            const chainProxy = chain.map(dep => {
+                const transformIds = this._registry.getClosestPlanTransformIds(dep.id, prevExec);
+                return EntryProxy.getFromEntry(dep, transformIds);
+            });
+            const result = currentGst.transform(context, chainProxy);
+            const [proxiedMain] = chainProxy;
+            proxiedMain.id = convertVariationalPath(
```
stephanwlee: This creates rsync like main file.

If base has a main that variation does not have, this creates the virtual entry. I might just add virtual entry instead later.

packages/mendel-pipeline/src/step/gst.js
```diff
@@ -0,0 +1,174 @@
...
+
+        // graph has array of arrays. First, make a pass and gather all variation info
+        const variations = new Set();
+        graph.forEach(nodes => {
+            nodes.forEach(node => {
+                variations.add(node.variation);
+            });
+        });
+
+        Array.from(variations.keys())
```
stephanwlee: This does not explore the permutation of variations. Instead, it creates one JSS for single variation only.

This is the part that requires good improvement with your help

examples/full-example/.mendelrc_v2
```diff
@@ -34,6 +34,8 @@ transforms: # a list of all available transforms for all envs and types
         - babili
   json-parse:
     plugin: mendel-parser-json
+  atomic-css:
+    plugin: mendel-atomic-css
```
irae: Well, I would prefer to not mix this into Mendel. It is kind of a line between proprietary and open source here.

packages/mendel-atomic-css/index.js
```diff
@@ -0,0 +1,74 @@
...
+    const entryMap = new Map();
+    entries.forEach(entry => entryMap.set(entry.normalizedId, entry));
+
+    return runEntryInVM(mainEntry, sandbox, (id) => entryMap.get(id));
+}
+
+module.exports = {
+    kind: 'gst',
+    predicate: function(entry) {
+        return !isNodeModule(entry.id) && entry.source.indexOf('getAtomicClasses') >= 0;
```
irae: Right now `isNodeModule` is OK, but should mendel itself only run this plugin against types that were assigned to this plugin on the mendelrc? Mono-repos and sub projects would benefit of the "node_module" restriction being removed.

packages/mendel-atomic-css/index.js
```diff
@@ -0,0 +1,74 @@
...
+    },
+    transform: function(context, entries) {
```
irae: For the sake of comparison, here is a snippet of a `mendel-transform` API:

```js
function({source, filename}, options) {
  /*__*/
  return {source, map};
}
```

How about we make GST API very similar?

```js
function({source, filename, deps}, options, context) {
  /*__*/
  return {source, map, deps};
}
```

The above suggestion would not change any principle. Just make it look similar.
But if you are willing to consider larger changes, I would go further and do something like:

```js
function({source, filename, deps}, options) {
  /*__*/
  return {source, map};
}
```

I think from our previous conversation, that you got the impression that I am just avoiding work. It is not the case, I think it is about making things better. Here is a bit of reasoning:

  1. Reduced API surface surface feels nice. `addVirtualEntry` pattern is not something I am super excited about, even in potential final form. I think it will be a hard concept to convey in documentation, requires too much understanding of Mendel internals and having and API like that is supporting edge cases that will happen in the future that we don't even foresee, people will use in many weird ways.
  2. `deps` in the return value is almost a side effect. It opens a can of worms: should we honor even if we don't find the dep? should we actually check? should we keep the previous values? are those deps just the direct dependencies or should it be the array with the whole graph of descendants (that's the input deps, so why output is objec)?

This would be so close to regular transform. Looks better, feels better, easy to explain: "If you need to look at your deps, you do GST, if not, it is IST." or even better, instead of `{type, predicate, transform}` it could be `{waitsForDeps=<Boolean>, predicate, transform}`.

Just think about it. We can leave as is, I am ok with it and I do understand the potential of future additions to context you are envisioning. I just wanted to make sure I give you the other side of the argument in full with a bit of better reasoning than our previous conversations.

packages/mendel-config/src/variation-config.js
```diff
@@ -10,10 +10,10 @@ const validate = createValidator({
 function VariationConfig(config) {
     const variations = parseVariations(config);
     const allVariationDirs = getAllDirs(variations);
-
     const baseVariation = {
         id: config.baseConfig.id,
         chain: [config.baseConfig.dir],
+        dir: config.baseConfig.dir
```
irae: I like this!

Thinking out loud: I think in the future we should do this for every variation and refactor `variationMatches` to have this instead. I am thinking about it too. Should we validate further making sure we have exactly one canonical dir per variation, even if we don't match the IDs? It is probably a nice idea.

packages/mendel-pipeline/src/step/gst.js
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * @param {MendelRegistry} tool.registry
+     * @param {DepsManager} tool.depsResolver
+     */
+    constructor({registry, depsResolver, cache}, options) {
+        super();
+
+        this._cache = cache;
+        this._registry = registry;
+        this._depsResolver = depsResolver;
```
irae: `depsResolver` not being used?

packages/mendel-pipeline/src/step/waiter.js
```diff
@@ -0,0 +1,20 @@
...
+        this.waitCount = 0;
+        this.cache = cache;
+    }
+
+    perform() {
+        this.waitCount++;
+        if (this.cache.size() > this.waitCount) return;
+        this.cache.entries().forEach(({id}) => this.emit('done', {entryId: id}));
+    }
+}
```
irae: That is a cleaver step implementation.

It creates a virtual bottleneck tho. For instance, before this step, the consumer/generator/outlet/middleware side cache was being wormed up in parallel.

`// TODO: Once type refactor is done and plan is part of entry instance, it will be safe to add optimization here to emit entries without GST right away before wait condition is met`

packages/mendel-pipeline/src/step/initialize.js
```diff
@@ -17,8 +16,8 @@ class Initialize extends BaseStep {
             .forEach(entry => this.pushEntryId(entry.id));
     }
 
-    pushEntryId(id) {
-        this.emit('done', {entryId: id});
+    pushEntryId(entryId) {
+        this.emit('done', {entryId});
```
irae: thanks! my bad!

packages/mendel-pipeline/src/step/gst.js
irae: Yeah, I can refactor later, we should pair on this so I share how I would do it.

irae:
[LGTM](https://lgtm.in/i/Qegg5iK6q)

[:+1:](https://lgtm.in/u/Qegg5iK6q)

Very nice overall! Think about the API and merge it once you are happy.

stephanwlee:
Thank you for the thorough review. Have a one last skim (removed all atomic-css related code) and merge away!

examples/full-example/package.json
```diff
@@ -38,6 +38,8 @@
     "karma-spec-reporter": "0.0.26",
     "mendel": "*",
     "mendel-development-middleware": "*",
+    "mendel-generator-extract": "*",
+    "mendel-generator-node-modules": "*",
```
irae: err, thanks, my bad again

PR #64, opened 2016-12-01
Cache per environment

stephanwlee:
This PR is attempting to remove complexity caused by cache/entries containing environment information. Please take a look at it. @irae

### DIFF
--- a/packages/mendel-pipeline/src/cache/server.js
+++ b/packages/mendel-pipeline/src/cache/server.js
@@ -10,7 +10,7 @@ const error = require('debug')('mendel:net:server:error');
 const verbose = require('debug')('verbose:mendel:net:server');
 
 class CacheServer extends EventEmitter {
-    constructor(config, cache) {
+    constructor(config, cacheManager) {
         super();
 
         this.config = config;
@@ -18,7 +18,7 @@ class CacheServer extends EventEmitter {
 
         this.clients = [];
 
-        this.cache = cache;
+        this.cacheManager = cacheManager;
         this.initCache();
 
         this.server = network.getServer(config.cachePort);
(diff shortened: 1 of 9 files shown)


yahoocla:
CLA is valid!

packages/mendel-pipeline/src/cache/server.js
```diff
@@ -63,30 +63,23 @@ class CacheServer extends EventEmitter {
...
 
     initCache() {
-        this.cache.on('doneEntry', entry => this.sendEntry(entry));
-        this.cache.on('entryRemoved', id => this.signalRemoval(id));
+        this.cacheManager.on('doneEntry', (cache, entry) => {
+            this.clients
+                .filter(client => client.environment === cache.environment)
+                .forEach(client => this._sendEntry(client, cache.size(), entry));
+        });
+        this.cacheManager.on('entryRemoved', id => this.signalRemoval(id));
```
irae: Should the signal removal also be per environment? I know most cases are because of the file being removed from the FS in the source folder. But it could also be a dependency removed by GST that only happens on environments the GST is on...  just in case, I guess.

packages/mendel-pipeline/src/daemon.js
```diff
@@ -15,17 +15,55 @@ require('./helpers/analytics/analytics-collector').setOptions({
...
+    }
+
+    removeEntry(id) {
+        Array.from(this._caches.values())
+            .forEach(cache => cache.removeEntry(id));
+        this.emit('entryRemoved', id);
+    }
+
+}
+
```
irae: This whole block could be a new file, right? Just to keep things tidy, idk

irae:
I like it, I receive all PRs as notification on this repo even if you don't @irae me 😉

irae:
I was about to branch from your branch but you were on a remote repo so I just merged it

PR #65, opened 2016-12-02
`normalizedId` now accounts for `package.json`

irae:
I feel like this is complete from the calculation side. I looked briefilly to the possibility of invalidating the files instead of throwing and added some inlide comments after I decided it is not time to do it.

There might be implications on what happens when we actually require `./components/about` on full example, as I didn't test this integration as of this commit. Once we do, there are implications for package.json, for instance, we have more than 1 `.components/about` for the `base` variation and we are not disambiguating.

Also, we can optimize the `deps` step to avoid filesystem sometimes. @stephanwlee did you notice how slow is deps in comparison to transforms now that GST is merged? It seems this optimization might be worthwhile after all.

Please, @stephanwlee please review carefully, I don't want to merge this yet without your feedback. Unless you think this is already good, in this case you can merge and work on top of it to avoid conflicts.

### DIFF
--- a/packages/mendel-pipeline/src/cache/entry.js
+++ b/packages/mendel-pipeline/src/cache/entry.js
@@ -5,6 +5,7 @@ class Entry {
         // property is filled by cache where they have more context
         this.normalizedId;
         this.variation;
+        this.runtime; // one of 'broser' or 'server'
 
         this.rawSource;
         this.source;
--- a/packages/mendel-pipeline/src/cache/entry.js
+++ b/packages/mendel-pipeline/src/cache/entry.js
@@ -5,7 +5,7 @@ class Entry {
         // property is filled by cache where they have more context
         this.normalizedId;
         this.variation;
-        this.runtime; // one of 'broser' or 'server'
+        this.runtime; // one of 'broser', 'server', 'package', 'isomorphic'
 
         this.rawSource;
         this.source;
--- a/packages/mendel-pipeline/src/cache/index.js
+++ b/packages/mendel-pipeline/src/cache/index.js
@@ -8,9 +8,12 @@ const variationMatches = require('mendel-development/variation-matches');
 class MendelCache extends EventEmitter {
     constructor(config) {
         super();
+        this.projectRoot = config.projectRoot;
         this.environment = config.environment;
         this._store = new Map();
         this._normalizedIdToEntryIds = new Map();
+        this._packageMap = new Map();
+        this._runtimeMap = new Map();
         this._baseConfig = config.baseConfig;
         this._variations = config.variationConfig.variations;
     }
@@ -29,6 +32,12 @@ class MendelCache extends EventEmitter {
             }
         }
 
+        if (this._packageMap.has(normalizedId)) {
+            const map = this._packageMap.get(normalizedId);
+            normalizedId = map.mapToId;
+            this._runtimeMap.set(id, map.runtime);
+        }
+
         return normalizedId;
     }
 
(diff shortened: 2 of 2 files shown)


yahoocla:
CLA is valid!

packages/mendel-pipeline/src/cache/index.js
```diff
@@ -39,21 +48,86 @@ class MendelCache extends EventEmitter {
...
     }
 
+    getRuntime(id) {
+        /about/.test(id) && console.log(id);
+        let runtime = 'isomorphic';
+        if (this._runtimeMap.has(id)) {
+            runtime = this._runtimeMap.get(id);
+        }
+        if (path.parse(id).base === 'package.json') runtime = 'package';
+        /about/.test(id) && console.log(runtime);
```
stephanwlee: remove me later on

packages/mendel-pipeline/src/cache/index.js
```diff
@@ -8,9 +8,12 @@ const variationMatches = require('mendel-development/variation-matches');
 class MendelCache extends EventEmitter {
     constructor(config) {
         super();
+        this.projectRoot = config.projectRoot;
         this.environment = config.environment;
         this._store = new Map();
         this._normalizedIdToEntryIds = new Map();
+        this._packageMap = new Map();
+        this._runtimeMap = new Map();
```
stephanwlee: Just wondering, how will this be used?

packages/mendel-pipeline/src/cache/entry.js
```diff
@@ -5,6 +5,7 @@ class Entry {
         // property is filled by cache where they have more context
         this.normalizedId;
         this.variation;
+        this.runtime; // one of 'broser' or 'server'
```
stephanwlee: or `package` or `isomorphic`.

I thought about this. At the manifest walking time, we need to pick the right one, right?

I will write down in my note that this is a limitation with Mendel: same file cannot be pointed from multiple package.json. A file cannot be both isomorphic and browser/main.

packages/mendel-pipeline/src/cache/index.js
```diff
@@ -39,21 +48,86 @@ class MendelCache extends EventEmitter {
...
+                this._normalizedIdToEntryIds.get(packageNormId).join('\n'),
+                '\n',
+            ].join(' '));
         }
+
+        delete require.cache[require.resolve(pkgPath)];
+        const pkg = require(pkgPath);
+
+        ['browser', 'main'].forEach(runtime => {
+            if (!pkg[runtime]) return;
```
stephanwlee: Filter!

packages/mendel-pipeline/src/cache/index.js
```diff
@@ -39,21 +48,86 @@ class MendelCache extends EventEmitter {
...
+            const targetNormId = this.getNormalizedId(
+                './' + path.join(parts.dir, pkg[runtime])
+            );
+            this._packageMap.set(targetNormId, {
```
stephanwlee: Can you add a small check? If `this._packageMap.has(targetNormId)`, throw saying multiple `package.json`s are pointing at the same file? Thanks!

packages/mendel-pipeline/src/cache/index.js
irae: WOW, I missed that... too much logs, I use `DEBUG=*mendel*` too much. Thanks

packages/mendel-pipeline/src/cache/entry.js
irae: I opened an issue for you to track constraints 😆 

Yeah... it is a bummer, but if we did all the huge job to support all combinations and edge cases, it would still be a good practice to organize your repo within the constraints. Makes things easier.

packages/mendel-pipeline/src/cache/index.js
irae: It will be useful for outlets. `cache.getEntriesByNormId('./components/about')` will return entries for both client and server. On the pipeline this don't meter, we just compile all of files anyway.

An outlet to write bundles will need this to write separate bundle for server and for client. The manifest on version v2 will contain both and on the server we get `main` and to serve bundles we get `browser`. Maybe `isomorphic` and `package` will never be really used, I left there mostly for debugging reasons. But maybe some crazy plugin actually wants to read more from the `package.json` than `browser` and `main` stuff.

packages/mendel-pipeline/src/cache/index.js
irae: good catch

irae:
@stephanwlee are we good now? can you double check?

stephanwlee:
👍

PR #84, opened 2017-03-06
Fix package dependency in examples

muralikr:
@stephanwlee Please review.

### DIFF
--- a/examples/Readme.mdown
+++ b/examples/Readme.mdown
@@ -2,9 +2,15 @@
 
 All examples are small applications that you can go to it's root directory and run `npm install` to have it all setup.
 
-You can then run:
+For Mendel 1.x, you can run:
 
+    $ npm run build
     $ npm run development
+
+For Mendel 2.x, you can run:
+
+    $ npm run buildv2
+    $ npm run v2
 
 To start the server in development mode. In this mode every source maps are enabled and bundle compilation is done on demand. The first time you load the application is slow, but once you save a file, the full bundle, and all the bundle combinations, should have changes propagated almost immediately.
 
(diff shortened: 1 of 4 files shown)


yahoocla:

examples/full-example/isomorphic/base/main.js
```diff
@@ -3,7 +3,6 @@
    See the accompanying LICENSE file for terms. */
 import App from './components/app';
 import React from 'react'; // eslint-disable-line
-import 'fake';
```
stephanwlee: hmm can you leave this? this is testing the "shim" concept. Yes, it doesn't work in Mendel v1.

examples/planout-example/Readme.md
```diff
@@ -78,6 +78,7 @@ To run this example go to it's root directory and run `npm install`.
 
 You can then run:
 
+    $ npm run build
```
stephanwlee: I don't think this is correct. This is running v1 version. Since this planout-example is covering only v1, I suppose it is correct but you should update the README of full-example to limit the scope.

stephanwlee: Can you update this guy for V2? https://github.com/yahoo/mendel/blob/master/examples/Readme.mdown

muralikr:
@stephanwlee Update PR.

examples/Readme.mdown
```diff
@@ -2,10 +2,16 @@
...
-You can then run:
+For Mendel 1.x, you can run:
 
+    $ npm run build
     $ npm run development
 
+For Mendel 2.x, you can run:
+
+    $ npm run buildv2
+    $ npm run v2
```
irae: @stephanwlee really? I suggested so many times to keep this just on the npm install stuff. having two ways feels kinda sloppy

PR #88, opened 2017-03-15
Adding More comments for .mendelrc in full example

muralikr:
@anuragdamle @stephanwlee @irae 

Added missing `repository` field in package.json files.

### DIFF
--- a/examples/full-example/.mendelrc
+++ b/examples/full-example/.mendelrc
@@ -1,7 +1,7 @@
 base-config:
-  id: base
-  dir: ./isomorphic/base
-  outdir: ./build
+  id: base # Bundle id for default variation
+  dir: ./isomorphic/base # Source dir for default variation
+  outdir: ./build # Base output dir for all artifacts
 
 variation-config:
   variation-dirs:
@@ -18,17 +18,19 @@ variation-config:
 
 route-config:
   # bundle does not have extension has it can be css bundle.
-  variation: /mendel/:variations/:bundle
-  hash: /mendel/:hash/:bundle
+  variation: /mendel/:variations/:bundle # Route pattern during development for variational assets
+  hash: /mendel/:hash/:bundle # Route pattern during production for variational assets
 
+# plugins are declared upfront and used by "types" bellow
+# you can have multiple instances of the same plugin with different ids
 transforms: # a list of all available transforms for all envs and types
-  babel:
-    plugin: mendel-transform-babel
-    options:
+  babel: # id or reference of a plugin instance
+    plugin: mendel-transform-babel # name of the plugin to be used, plugin must be installed in your project
+    options: # configuration object passed to plugin instance, each plugin has it's own format and documentation
       presets:
         - es2015
         - react
-  uglyfy: # lazy to write the real one, will be faster if we have real uglify?
+  uglyfy: # Example configuration for Babel + Uglification
     plugin: mendel-transform-babel
     options:
       comments: false # true - Output comments in generated output.
@@ -37,13 +39,13 @@ transforms: # a list of all available transforms for all envs and types
         -
           - es2015
           -
-            loose: true
+            loose: true # Preset "es2015-loose" is incompatible for mendel
         - react
-  json-parse:
+  json-parse: # Parsing JSON files to be usable in JS
     plugin: mendel-parser-json
   less-transform:
     plugin: mendel-transform-less
-  babel-test:
+  babel-test: # Example configuration for istanbul
     plugin: mendel-transform-babel
     options:
       presets:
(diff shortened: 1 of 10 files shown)


anuragdamle:
👍

examples/full-example/.mendelrc
```diff
@@ -18,17 +18,17 @@ variation-config:
...
-  hash: /mendel/:hash/:bundle
+  variation: /mendel/:variations/:bundle # Route pattern during development for variational assets
+  hash: /mendel/:hash/:bundle # Route pattern during production for variational assets
 
 transforms: # a list of all available transforms for all envs and types
   babel:
-    plugin: mendel-transform-babel
-    options:
+    plugin: mendel-transform-babel # Name of package for transformations
+    options: # Extra arguments to be passed to plugin
```
irae: I think we should document only the first plugin, to remove clutter, and here is a suggestion:
```yaml
  # plugins are declared upfront and used by "types" bellow
  # you can have multiple instances of the same plugin with different ids
  babel:                            # id or reference of a plugin instance
    plugin: mendel-transform-babel  # name of the plugin to be used, plugin must be installed in your project
    options:                        # configuration object passed to plugin instance, each plugin has it's own format and documentation
```

examples/full-example/.mendelrc
```diff
@@ -53,23 +53,19 @@ transforms: # a list of all available transforms for all envs and types
...
       - .json
-    parser: json-parse
-    parser-to-type: js
-    include-node_modules: true
+    parser: json-parse # Uses parser defined in transforms.
+    parser-to-type: js # Maps to type defined below
+    include-node_modules: true # Include files matching node_modules/* path.
   js:
-    outlet:
-      plugin: mendel-bundle-browser-pack
```
irae: Why is outlet being removed as part of a documentation PR? Did the spec change and this entries were forgotten?

examples/full-example/.mendelrc
```diff
@@ -79,15 +75,15 @@ types:
       - .less
     parser: less-transform
     parser-to-type: css
-    resource: true
+    resource: true #
```
irae: trailing comment is empty

examples/full-example/.mendelrc
```diff
@@ -53,23 +53,19 @@ transforms: # a list of all available transforms for all envs and types
         - istanbul
 
 types:
-  json:
+  json: # Type id. Doesn't have to be file extension.
```
irae: Same suggestion as `transforms`. One or two lines before "types:" explaining why types exist, and document only first one with column aligned comments.

examples/full-example/.mendelrc
```diff
@@ -53,23 +53,19 @@ transforms: # a list of all available transforms for all envs and types
...
     extensions:
       - .json
-    parser: json-parse
-    parser-to-type: js
-    include-node_modules: true
+    parser: json-parse # Uses parser defined in transforms.
+    parser-to-type: js # Maps to type defined below
```
irae: AFAIK `parser:` is optional, and `parser-to-type:` is likely mandatory only when `parser:` is present. Comments should say that.

examples/full-example/.mendelrc
```diff
@@ -79,15 +75,15 @@ types:
...
     parser-to-type: css
-    resource: true
+    resource: true #
   css:
     extensions:
       - .css
     resource: true
 
-outlets:
+outlets: # Defines the way we output bundles
```
irae: Same suggestion about comment format, similar to `transforms` and `types`.

examples/full-example/.mendelrc
```diff
@@ -210,4 +209,6 @@ env:
         outfile: server.js
         entries:
           - ./main.js
-        runtime: main # This
+        # If npm package has multiple entry files, this pulls only
+        # server side files defined in "main" field of package.json
+        runtime: main
```
irae: This is pretty hard to understand. I don't fully follow to be honest, although I do understand the concept of multiple runtimes package.json very well.

Maybe say something "`runtime` forces a particular bundle to a certain runtime, per CommonJS spec. See more at:" and find an useful link or write a new documentation on why this exists in Mendel.

The core suggestion here is: It is too much to explain on the full example, use something simple (maybe even simpler than my suggestion) and link to a full spec or full docs.

examples/full-example/.mendelrc
stephanwlee: sounds like a good idea! thanks for the feedback

examples/full-example/.mendelrc
stephanwlee: the implementation is not using this at all. Frankly, I am still contemplating about this but since we are not honoring this config, I thought it would be a good idea before people start to copy it.

See "bundles" having the outlet property. We can talk about this offline. I need some guidance.

examples/full-example/.mendelrc
stephanwlee: Frankly, this has to do with implementation of one of the outlet so it should go into respective outlets documentation. Thanks for the feedback. Maybe it is a better idea to write better documentation on auxiliary packages and put a link to those in this readme.

irae: I think this is a remaining design flaw from my initial mendel-v2-rc draft. Think about it: All transforms are per instance config, all types are per "id" config. All generators and outlets should be the same. You should reference and outlet or generator per ID of it's instance and not have a generator perform differently based on the bundle config.

This is a good candidate for a small revision on the `.mendelrc` spec. For now, we could follow example of the other docs, declaring a bundle and what properties are bundle specific and mendel specific, than we could say: "other properties are used by either a generator or an outlet, please refer to ....".

muralikr:
@irae Updated comments as per your suggestion.

packages/mendel-outlet-server-side-render/package.json
```diff
@@ -8,6 +8,10 @@
   },
   "author": "",
   "license": "ISC",
```
stephanwlee: can you fix me? :(

PR #90, opened 2017-03-28
Add bundle config warning for unsupported props

muralikr:
@anuragdamle @irae @stephanwlee

### DIFF
--- a/packages/mendel-config/src/bundle-config.js
+++ b/packages/mendel-config/src/bundle-config.js
@@ -7,6 +7,7 @@ function BundleConfig(id, options, config) {
     this.generator = options.generator || 'default';
 
     const {outdir} = config.baseConfig;
+
     this.outfile = options.outfile ?
         path.resolve(outdir, options.outfile) : '';
 
@@ -23,6 +24,7 @@ function BundleConfig(id, options, config) {
         options,
         ['generator', 'outfile', 'outlet', 'entries', 'require', 'external']
     );
+
     this.options = undash(this.options);
 
     BundleConfig.validate(this);
@@ -30,6 +32,19 @@ function BundleConfig(id, options, config) {
 
 BundleConfig.validate = createValidator({
     id: {required: true},
+    supportedOptionFields: [
+        'allBundles',
+        'entries',
+        'external',
+        'from',
+        'generator',
+        'manifest',
+        'outfile',
+        'outlet',
+        'require',
+        'runtime',
+        'sourcemap',
+    ],
 });
 
 function flattenArrays(inArray) {
--- a/packages/mendel-config/src/validator.js
+++ b/packages/mendel-config/src/validator.js
@@ -25,6 +25,16 @@ module.exports = function createValidator(schema) {
             }
         });
 
+        if (instance.options && Array.isArray(schema.supportedOptionFields)) {
+            const unsupportedOptionFields = Object.keys(instance.options).filter(field => {
+                return schema.supportedOptionFields.indexOf(field) === -1;
+            });
+
+            if (unsupportedOptionFields.length > 0) {
+                error.push('Found unsupported options `' + unsupportedOptionFields.join('`, `') + '`');
+            }
+        }
+
         if (error.length) {
             throw new Error(
                 error.filter(Boolean).reduce(function(reduced, error) {


anuragdamle:
👍

examples/full-example/.mendelrc
```diff
@@ -17,8 +17,8 @@ variation-config:
       - partner_C
 
 route-config: # Mendel-middleware related
-  variation: /mendel/:variations/:bundle # Route pattern for variational assets during development
-  hash: /mendel/:hash/:bundle # Route pattern for variational assets during production
+  variation: /mendel/:variations/:bundle.js # Route pattern for variational assets during development
```
stephanwlee: I did not put the extension on purpose. Check the example app and see url of the payloads. More specifically, note the css bundle. Css bundle is not js.

packages/mendel-config/src/bundle-config.js
```diff
@@ -23,6 +36,20 @@ function BundleConfig(id, options, config) {
...
+    if (unsupportedOptions.length > 0) {
+        console.log([
+            '[Config][WARN] Following options are not supported for bundle: ' + id + '.\n',
+            unsupportedOptions.join(', '),
+        ].join(' '));
+    }
+
     this.options = undash(this.options);
 
     BundleConfig.validate(this);
```
stephanwlee: Can enum be a feature of the validator instead?

examples/full-example/.mendelrc
irae: I think adding extension should be a job for the outlet. Agreed it should not be part of configuration. pathToRegexp sometimes breaks with .js, since . means any character. Express works around it to make config easier, but we don't use express, we use pathToRegex directly, right? Lets avoid all this complexity.

packages/mendel-config/src/bundle-config.js
```diff
@@ -7,6 +7,19 @@ function BundleConfig(id, options, config) {
...
+        'entries',
+        'external',
+        'from',
+        'generator',
+        'manifest',
+        'outfile',
+        'outlet',
+        'require',
+        'runtime',
+    ];
```
irae: I think it would be better to infer from basic config. Actually, I thought the code already did that. I remember writing routes for picking only properties that existed on the default config from the parsed config. Maybe the code was removed. Either way, listing manually makes hard to explore and develop new features. Please get the list of supported options from `Object.keys(defaultConfig)`.

examples/full-example/.mendelrc
stephanwlee: Outlet simply transform "bundle" object into something. This route business is an applicational concept that does not belong in outlet. Yeah i wondered why we dont use express. 

Random thought: multiple mendel clients per express app feels bad. :(

packages/mendel-config/src/bundle-config.js
```diff
@@ -23,6 +36,20 @@ function BundleConfig(id, options, config) {
...
+
+    Object.keys(this.options).forEach((optionKey) => {
+        if (supportedOptions.indexOf(optionKey) === -1) {
+            unsupportedOptions.push(optionKey);
+        }
+    });
+
+    if (unsupportedOptions.length > 0) {
+        console.log([
+            '[Config][WARN] Following options are not supported for bundle: ' + id + '.\n',
```
irae: Please use `require('debug')('mendel:config')`. This warning is non essential in most scenarios, since you can have YAML pointers to objects of arbitrary names the user might create.

packages/mendel-config/src/bundle-config.js
muralikr: This warning is for extra props defined in each `bundle`. `Object.keys(defaultConfig)` gives only top level keys of `.mendelrc`

stephanwlee: The code you are referring to is done on first level keys. Perhaps we can apply that recursively while plugin specific options can be swept under "options" key that doesnt have any validation or removal.

muralikr: > The code you are referring to is done on first level keys.

`this.options` is updated in this (line https://github.com/yahoo/mendel/pull/90/files#diff-1f6220fbe492ebe47f6ff935346291aeL22).

After this, `this.options` will only have config defined for given bundle.

packages/mendel-config/src/bundle-config.js
muralikr: Didn't get you, can you please elaborate?

examples/full-example/.mendelrc
```diff
@@ -130,7 +130,7 @@ bundles: # Definition of a bundle
   main:
     outlet: browser-pack
     outfile: main.js
-    entries:
```
stephanwlee: is this line correct?

packages/mendel-config/src/bundle-config.js
stephanwlee: There is a validator and this is the line that invokes the validation. Can whitelist of keys be enforced in the validator instead of having sprinkle of config here and there? (this is a recommendation and if you think it sucks, just ignore this suggestion)

packages/mendel-config/src/bundle-config.js
irae: OK, this leaves most of my comment as invalid. Either way, different generators will implement custom properties IIRC.

packages/mendel-config/src/bundle-config.js
```diff
@@ -19,10 +33,26 @@ function BundleConfig(id, options, config) {
...
+
+    Object.keys(this.options).forEach((optionKey) => {
+        if (supportedOptions.indexOf(optionKey) === -1) {
+            unsupportedOptions.push(optionKey);
+        }
+    });
+
+    if (unsupportedOptions.length > 0) {
+        debug([
+            '[Config][WARN] Following options are not supported for bundle: ' + id + '.\n',
```
irae: If you try out `DEBUG=mendel* mendel` the output already says in color that it is part of "mendel:config". So I think it might be better to remove the `[Config]` portion of the string.

packages/mendel-config/src/bundle-config.js
irae: @stephanwlee this is similar to my idea of "instances of generators" the same way we do "instances of plugins".

irae:
Weird, I thought I had a review going on, could not click approve. Regardless, once we have the right answer for where in the code the array of whitelisted props would live, you can merge.

[LGTM](https://lgtm.in/i/SNzJprUbq)

[:+1:](https://lgtm.in/u/SNzJprUbq)

muralikr:
@stephanwlee Moved config validation to validator module.

irae:
Wow! This looks way better! =)

packages/mendel-config/src/validator.js
```diff
@@ -25,6 +25,16 @@ module.exports = function createValidator(schema) {
             }
         });
 
+        if (Array.isArray(schema.supportedOptionFields)) {
+            const unsupportedOptionFields = Object.keys(instance.options).filter(field => {
```
stephanwlee: Just note that this generic code can have missing field "options"

PR #93, opened 2017-04-10
Update package dependencies of full-example

muralikr:
For `full-example`, `npm install` is not installing all latest mendel-* beta packages. I have updated `package.json` with minimum version required.

Also, we don't need `mendel` as devDependency in V2. Removed this package.

After this, `npm install` should install all required packages for `full-example`.

cc: @anuragdamle


irae:
👎 

I was trying to fix in a different way. Let me explain some principles before we try to solve this.

  1. All examples need to import any mendel packages with `"mendel-[name]": "*"` and the reason is local development. You can't `npm install` from examples because of it being in beta. In theory all examples need to work with stable non beta packages. So, with `npm run linkall` it should work even with the beta. for some reason it does not.
  2. We must have a `package.json` that uses `mendel`. When I changed root `package.json` to `mendel-monorepo` the idea was to have `mendel-cli` with `package.json` with `{"name":"mendel"}`. This is likely a good way to move forward, but we need to decouple CLI from `mendel-pipeline`, maybe rename it to `mendel-daemon` so it matches most of it's functionality. This is a slightly larger refactor, but probably worth it.
3. There is too much dead code in many places. Examples, `mendel-config`, `mendel-requirefy` itself is likely not used. I rather we have kept all of mendelV1 packages, but `mendel-browserify` is gone and `mendel-requirefy` and other modules are still there. We need a full cleanup in addition to just a quick fix for full-example.

muralikr:
> 1. All examples need to import any mendel packages with "mendel-[name]": "*" and the reason is local development. You can't npm install from examples because of it being in beta. In theory all examples need to work with stable non beta packages. So, with npm run linkall it should work even with the beta. for some reason it does not.

I have changed it back to "*". npm install is not pulling latest packages from registry. After a clean install, this is the output of `npm outdated`. I see the same happening in other repos too.

```
Package                                              Current  Wanted        Latest  Location
babel-plugin-istanbul                                  3.1.2   3.1.2         4.1.1  mendel-full-example
babel-plugin-transform-inline-environment-variables    6.8.0   6.8.0         0.0.2  mendel-full-example
babel-preset-babili                                    0.0.8   0.0.8        0.0.12  mendel-full-example
envify                                                 3.4.1   3.4.1         4.0.0  mendel-full-example
eslint-plugin-react                                    5.2.2   5.2.2        6.10.3  mendel-full-example
karma-jasmine-diff-reporter                            0.6.3   0.6.3         1.0.0  mendel-full-example
karma-spec-reporter                                   0.0.26  0.0.26        0.0.30  mendel-full-example
mendel-development-middleware                          2.1.1   2.1.1  3.0.0-beta.2  mendel-full-example
mendel-manifest-uglify                                 1.0.0   1.0.0  3.0.0-beta.2  mendel-full-example
mendel-middleware                                      2.1.2   2.1.2  3.0.0-beta.2  mendel-full-example
mocha                                                  2.5.3   2.5.3         3.2.0  mendel-full-example
```
`full-example` requires latest versions of mendel packages. So, it is failing with above versions. npm is installing latest version only by forcing version number.


> 2. We must have a package.json that uses mendel. When I changed root package.json to mendel-monorepo the idea was to have mendel-cli with package.json with {"name":"mendel"}. This is likely a good way to move forward, but we need to decouple CLI from mendel-pipeline, maybe rename it to mendel-daemon so it matches most of it's functionality. This is a slightly larger refactor, but probably worth it.

@stephanwlee: since you have more context on this, can you add your inputs on removing `mendel` from `dependencies` ?? 

> 3. There is too much dead code in many places. Examples, mendel-config, mendel-requirefy itself is likely not used. I rather we have kept all of mendelV1 packages, but mendel-browserify is gone and mendel-requirefy and other modules are still there. We need a full cleanup in addition to just a quick fix for full-example.

Agree that we have lot of dead code in many places. We have to clearly separate V1 and V2 dependencies and remove dead code. We have this task in our timeline. Feel free to create a PR if you get some free time.

irae:
About of 2) I think the best way would be to split `mendel-pipeline` into `mendel-cli` and `mendel-daemon`. I don't know if the build part is still coupled, but we can also extract it apart, but I wouldn't plan upfront. I think CLI might be a good place for the build code. Only after the first split we could make a right call on that.

Then, `mendel-cli`should have `{name:"mendel"}` on `package.json`.

About `*`: Just remove the "beta" tag and mark it as final, than it will work again, both with `npm run link` and with `npm install`.

It is not my call to say it is final, but if Norrin uses it in production, my personal opinion is that it is already final 😁

stephanwlee:
Mendel is removed in favor of mendel-pipeline which now exposes bin of
"mendel".

For full example, you should use npm run linkall, not npm i.

I am reading between lines so I may have answered wrongly.

irae:
Yes, @stephanwlee, this is the current state, I was talking about the direction I think we should take it.

`mendel-pipeline` publishing `mendel` is a bit messy, and we don't have a central place for CLI. We need CLI to scale up with more commands like variation-diff, bootstrap, merge, etc.

examples/full-example/package.json
```diff
@@ -60,10 +59,12 @@
     "mendel-pipeline": "*",
     "mendel-transform-babel": "*",
     "mendel-transform-less": "*",
+    "mocha": "^2.5.0",
     "mocha-istanbul-reporter": "^0.1.0",
     "nodemon": "^1.9.2",
     "react-addons-test-utils": "^15.4.2",
-    "rtlcss": "^2.1.2"
+    "rtlcss": "^2.1.2",
+    "watchify": "^3.0.0"
```
irae: I was under the impression we didn't use `watchify` anymore. leftover?

irae:
I just tested this with a clean repo, or rather, with `git clean -dfx`. Works well, but there is a problem. NPM chokes if there is already mendel installed at the global scope or previously linked somewhere.

```
mendel-pipeline irae$ npm link
npm WARN gentlyRm not removing /Users/irae/.nvm/versions/node/v6.9.1/bin/mendel as it wasn't installed by /Users/irae/code/mendel/packages/mendel-pipeline
```

It was only solved by using `--force` when linking from the `mendel-pipeline` directory.

I think you should merge, but next step would be to refactor as I suggested, making mendel-cli own the "mendel" package, this way the migration path happens only once.

irae:
[LGTM](https://lgtm.in/i/67kUQ5e9q)

[:+1:](https://lgtm.in/u/67kUQ5e9q) 

(please take a look at watchify comment)

examples/full-example/package.json
muralikr: It was a dependency for `mendel-development-middleware@2.1.1`. I have removed it now, as it is not required for `mendel-development-middleware@3.0.0-beta.2`.

irae:
👍 thanks

PR #97, opened 2017-05-09
Use sha-sum instead of integers in prune

muralikr:
As of now, we are using incremented ids in prune post generator.
Order of files cannot be relied on and it causes unexpected
dependecy issues between multiple bundles. This is more vulnerable
when files are added or removed.

We should always create unique hash based on a file name. Irrespective of file order,
prune id should be unique.

@anuragdamle @irae @stephanwlee

### DIFF
--- a/packages/mendel-generator-prune/index.js
+++ b/packages/mendel-generator-prune/index.js
@@ -1,4 +1,5 @@
 const verbose = require('debug')('verbose:mendel:generator:prune');
+const shasum = require('shasum');
 
 // First argument is not needed; just to make it look like normal generator API
 function pruneModules(bundle, doneBundles, registry, generator) {
@@ -12,11 +13,6 @@ function pruneModules(bundle, doneBundles, registry, generator) {
 }
 
 function pruneGroup(bundles) {
-    const getUniqId = (function() {
-        let index = 0;
-        return () => '_u' + Number(index++).toString(16);
-    })();
-
     const exposeNormToExposeId = new Map();
     const allNorms = new Set();
 
@@ -25,8 +21,9 @@ function pruneGroup(bundles) {
             const norm = entry.normalizedId;
             allNorms.add(norm);
 
-            if (entry.expose && !exposeNormToExposeId.has(entry.expose))
-                exposeNormToExposeId.set(entry.expose, getUniqId());
+            if (entry.expose && !exposeNormToExposeId.has(entry.expose)) {
+                exposeNormToExposeId.set(entry.expose, shasum(norm));
+            }
         });
     });
 
@@ -57,8 +54,10 @@ function pruneGroup(bundles) {
             });
 
             // Remap externals
-            if (exposeNormToExposeId.has(entry.expose))
+            if (exposeNormToExposeId.has(entry.expose)) {
                 clone.expose = exposeNormToExposeId.get(entry.expose);
+            }
+
             Object.keys(clone.deps).forEach(literal => {
                 const deps = clone.deps[literal];
                 Object.keys(deps).forEach(runtime => {
--- a/packages/mendel-generator-prune/package.json
+++ b/packages/mendel-generator-prune/package.json
@@ -11,6 +11,7 @@
     "url": "https://github.com/yahoo/mendel"
   },
   "dependencies": {
-    "debug": "^2.6.3"
+    "debug": "^2.6.3",
+    "shasum": "^1.0.2"
   }
 }


irae:
Comment on the hashing algorithm below, but before that: Can you clarify why we need hashes?

The counter should be *internal* to the bundle, every file that can be accessed across bundles are exposed instead. Browserify uses a counter in the same way for the longest while. We've been using counters on Mendel v1 too for a while. It feels to me we might have a bug on the way we expose modules between bundles in Mendel v2 and hashing would hide the problem instead of solving the real issue.

Could you please provide an example of collision?  I understand the cross-bundle issue you mention, I just don't understand how this could happen if we are exposing only normalizedIds as we used to do. Maybe I am missing something and expose algo changed from v1 to v2?

Remember, the purpose of this generator is to make the bundle smaller. Adding a hash will increase bundle payloads, specially in codebases such as Yahoo with so many files.

So, code-wise I think this is good. I don't want to block the change or anything. I am just trying to be helpful here.

If we are going to use hashes I would say we should use shasum for now (since the remainder of the codebase is already using it). I was actually researching better hashes but before we migrate to a different algorithm I think it is safe to keep uniformity across. I am not sure this generator gets it, but we do have the shasum on the manifest, so we can reuse the sha generation so build times don't increase that much.

irae:
Also, would be nice to have a readme for this package with a bit of explanation on the role this generations plays in the architecture, maybe @stephanwlee would be the one to contribute a couple of paragraphs? 😉

anuragdamle:
@irae let me try to explain the issue.

Assume that we have 2 bundles. First bundle(the main bundle) has 3 files 

"foo.js":"_u1",
"bar".js":"_u2",
"foobar".js":"_u3",

There is a second bundle(lazy) which has its own files as well as files from main bundle.


"bar1.js":"_ua1",
"bar2.js":"_ua2",
"foobar".js":"_u3",

This is all good and works till this part.

Now Suppose that we add 2 more files in the main bundle. Now the file names and prune id are as follows

"foo.js":"_u1",
"bar".js":"_u2",
"barzz.js":"_u3",
"fooba""_u4",
"foobar".js":"_u5"

Note that the foobar.js prune id is changed to _u5 now. And since there is no change to the lazy bundle at all there is no update to that bundle. So the reference of foobar.js is _u3 in the lazy bundle. This leads to inconsistencies the problems that murali outlined.

irae:
Yes, I do understand that. This was explained in short version by @muralikr on the description. What I don't understand is why the lazy bundle is even referencing by id?

Also, even if it is, why should it meter? Don't you deploy manifest with both bundles all the time? shouldn't the IDs change with the bundles if that's the case?

For instance, to follow your example. When a module is exposed, this should happen:

"foo.js":"_u1",
"bar".js":"_u2",
"foobar".js":"/components/fobar",

And on lazy, which uses files from main bundle you should have:

"bar1.js":"_ua1",
"bar2.js":"_ua2",
"foobar".js":"/components/fobar",

When main bundle starts having more dependencies this should happen:

"foo.js":"_u1",
"bar".js":"_u2",
"barzz.js":"_u3",
"fooba""_u4",
"foobar".js":"/components/fobar"

Therefore the exposed module should never be affected by number of files within.

In other words: The bug is on how Mendel v2 exposes and not in generating a hash comparison. By doing a hash you will make build times slower, will also increase the payload of all bundles, and the expose issue will be hidden, and can backfire again in the future.

muralikr:
> The counter should be internal to the bundle, every file that can be accessed across bundles are exposed instead. Browserify uses a counter in the same way for the longest while. We've been using counters on Mendel v1 too for a while. It feels to me we might have a bug on the way we expose modules between bundles in Mendel v2 and hashing would hide the problem instead of solving the real issue.

Agree that file accessed in multiple bundles shouldn't have counter based id. There seems to be a bug on Mendel V2 in exposing these files. I will try to investigate that issue and update this PR.

> If we are going to use hashes I would say we should use shasum for now (since the remainder of the codebase is already using it). I was actually researching better hashes but before we migrate to a different algorithm I think it is safe to keep uniformity across. I am not sure this generator gets it, but we do have the shasum on the manifest, so we can reuse the sha generation so build times don't increase that much.

I have checked `shasum` package. It's hash has 32 characters which increases bundle payload. `hash-sum` returns hash of 8 characters (similar to git commit hashing).

irae:
> I have checked shasum package. It's hash has 32 characters which increases bundle payload. hash-sum returns hash of 8 characters (similar to git commit hashing).

There is a reason to use a good hash instead of any hash. When you use the hash you chose you might have collisions, meaning different files with the same hash. If this happens stuff will blow up. Thats the reason I chose sha1 over anything else at the time. Use worst algorithms at your own risk :wink:

muralikr:
@irae Updated code to use `shasum` algo. Thanks.

anuragdamle:
👍

irae:
Hey, when you push it, can you make it 2.0 final, please?

irae:
I mean, all packages to be 2.0 final?

PR #98, opened 2017-05-10
Updating package versions to 3.0.0

muralikr:
@irae @anuragdamle

### DIFF
--- a/packages/mendel-pipeline/package.json
+++ b/packages/mendel-pipeline/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mendel-pipeline",
-  "version": "3.0.0-beta.2",
+  "version": "3.0.0",
   "description": "The mendel package",
   "main": "src/main.js",
   "bin": {
@@ -22,10 +22,10 @@
     "cli-table": "^0.3.1",
     "commander": "^2.9.0",
     "debug": "^2.2.0",
-    "mendel-config": "^3.0.0-beta.2",
-    "mendel-deps": "3.0.0-beta.2",
+    "mendel-config": "^3.0.0",
+    "mendel-deps": "3.0.0",
     "mendel-development": "^1.1.0",
-    "mendel-resolver": "3.0.0-beta.2",
+    "mendel-resolver": "3.0.0",
     "node-libs-browser": "^2.0.0",
     "pretty-ms": "^2.1.0",
     "resolve": "^1.1.7"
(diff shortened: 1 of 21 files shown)


anuragdamle:
👍

packages/mendel-pipeline/package.json
```diff
@@ -22,10 +22,10 @@
     "cli-table": "^0.3.1",
     "commander": "^2.9.0",
     "debug": "^2.2.0",
-    "mendel-config": "^3.0.0-beta.2",
-    "mendel-deps": "3.0.0-beta.2",
```
irae: can this break for people? are you using internally version 3?

muralikr: We are using beta version internally. But, this is with beta tag. New packages will be created with 2.0.0 stable tag.

irae:
I forgot we need to fix `mendel-pipeline` and the executable problem.

Suggestion: Make "bin" for mendel-pipeline to be "mendel-pipeline" and have the main package "mendel" to have a executable that just exec mendel-pipeline internally. So root package.json would have only 1 dependency, the mendel-pipeline. and cli.js would just require mendel-pipeline.

packages/mendel-pipeline/package.json
irae: I think there is one package that used to be 2.0 before, I can't remember if this colides. Maybe mendel-core? you can `npm info` the ones that didn't change name and see if 2.0 will work or not.

Maybe we need all of them to be 3.0 because no package ever did become so high.

muralikr: Thanks for catching it. Updated version to 3.0.0

muralikr:
> have the main package "mendel" to have a executable that just exec mendel-pipeline internally. 

`mendel` package doesn't have `mendel-pipeline` dependency. So, we can't execute `mendel-pipeline` from this package bin. Are you referring to another package?

irae:
😆 

No! I am suggesting you do it!

Step 1: Change all dependencies there are there to no dependency at all, and add just `mendel-pipeline`.
Step 2: change `mendel-pipeline` package.json to `{"bin": {"mendel-pipeline": "..."}}`
Step 3: Create `{"bin": {"mendel": "bin/cli.js"}}` on `package.json` and create the `cli` that just reuses `mendel-pipeline` under the hood.

My explanation was bad. Sorry.

irae:
Let's discuss over Slack if you have further questions 😉

muralikr:
We are blocked on the prune change, so don't want to complicate things at this point. We can add this in a follow up PR.

irae:
It will be a mess if you push as is to `npm`. It means to fix it you will need to push everyone to `4.0.0` so you don't break stuff for everyone.

PR #99, opened 2017-05-12
Remove shasum and add Readme for generator-prune

muralikr:
@irae @anuragdamle @stephanwlee

### DIFF
--- a/packages/mendel-generator-prune/README.md
+++ b/packages/mendel-generator-prune/README.md
@@ -1,0 +1,24 @@
+# mendel-generator-prune
+
+This generator removes dangling files, such as files required initially and later not used because of some transformation (examples would be `requires` removed after `envify` and some CSS-in-JS dependency. It also assigns `NORMALIZED_ID` to files exposed in other bundles. `NORMALIZED_ID` is the relative path to either base or any variation.
+
+## Example
+Assume that we have 2 bundles. First bundle(the main bundle) has 3 files
+
+```
+"foo.js":"1",
+"bar".js":"2",
+"foobar".js":"NORMALIZED_ID",
+```
+
+There is a second bundle(lazy) which has its own files as well as files from main bundle as dependencies.
+
+```
+"deps": {
+    "dep1.js":"1",
+    "dep2.js":"2",
+    "foobar".js":"NORMALIZED_ID"
+}
+```
+
+Now, even if you update other files in `main` bundle and recreate this bundle, lazy bundle will always refer to `foobar.js` by it's NORMALIZED_ID.
--- a/packages/mendel-generator-prune/package.json
+++ b/packages/mendel-generator-prune/package.json
@@ -1,7 +1,7 @@
 {
   "name": "mendel-generator-prune",
-  "version": "3.0.0",
-  "description": "Mendel post-generator that cleans up dangling dependencies and mangles `expose` for shorter names.",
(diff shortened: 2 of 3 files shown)


anuragdamle:
👍

packages/mendel-generator-prune/README.md
```diff
@@ -0,0 +1,30 @@
+# mendel-generator-prune
+
+Mendel-generator modifies references of exposed modules across bundles.
+
+In a large application with lot of files, few common files are required by most of other files in other bundles. This generator assigns unique hash to each exposed module. Instead of referring those common files by file path, they will be referred by unique hash. Hash is computed using `shasum`.
```
irae: This is not exactly what it does.

Before:
1) It did remove dangling files, such as files required initially and later not used because of some transformation (examples would be `requires` removed after `envify` and some CSS-in-JS dependency.
2) Make sure we replace `normalizedId` (which is relative to variation and is already unique) by a counter, so that the files become really smaller.

After `shasum`: 
1) Still the same as above
2) Replace all file names by something else that still ensures cross bundle consistency. (same thing normalizedId already does?)

Now that I I am thinking here. Have you guys tried just not using `shasum` and not replacing by any numbers? just `normalizedId` and nothing else? Might actually solve the issue. If we remove the part of the code that renames the entries/entrances we can just keep reason 1 explained on the readme. To be honest, I think `normalizedIds` will compress better. Drawback is: they are readable by user that view source.

packages/mendel-generator-prune/README.md
```diff
@@ -0,0 +1,30 @@
...
+    "dep2.js":"2",
+    "foobar".js":"SOME_HASH"
+}
+```
+
+Now, even if you update other files in `main` bundle and recreate this bundle, lazy bundle will always refer to `foobar.js` by hash.
+
+#### Known issues
+
+`shasum` generates 32 character hash which does not have good compression rate with gzip. So, you will see an increase in your bundle payload.
```
irae: One more reason to try to remove it. Small hash might have collision, but no hash at all will not have collisions because `normalizedId` is already consistent.

muralikr:
@irae Removed `shasum` and updated Readme. Please review.

packages/mendel-generator-prune/README.md
```diff
@@ -0,0 +1,24 @@
+# mendel-generator-prune
+
+This generator removes dangling files, such as files required initially and later not used because of some transformation (examples would be `requires` removed after `envify` and some CSS-in-JS dependency. It also assigns `NORMALIZED_ID` to files exposed in other bundles.
```
irae: Suggestion here: `NORMALIZED_ID` is the relative path to either `base` or any variation.

packages/mendel-generator-prune/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mendel-generator-prune",
-  "version": "3.0.0",
-  "description": "Mendel post-generator that cleans up dangling dependencies and mangles `expose` for shorter names.",
+  "version": "3.0.1",
+  "description": "Mendel post-generator that cleans up dangling dependencies and normalizes `expose` fields.",
```
irae: suggestion to not use `expose` because this is a browserify term. We don't expose, we just never hide it anyway 😄

irae:
Nice! 👍

PR #101, opened 2017-05-15
Return 500 and Log error when bundle is empty

muralikr:
Sometimes, empty bundle is returned from mendel middleware. This gets cached in cdn or cache layers and served to more number of users.

We are still debugging the actual root cause of this issue, I have added logs to find no. of occurrences of this issue.

@anuragdamle @irae

### DIFF
--- a/packages/mendel-middleware/index.js
+++ b/packages/mendel-middleware/index.js
@@ -6,6 +6,7 @@ var pathToRegexp = require('path-to-regexp');
 var bpack = require('browser-pack');
 var MendelTrees = require('mendel-core');
 var MendelLoader = require('mendel-loader');
+var debug = require('debug')('mendel:middleware');
 
 module.exports = MendelMiddleware;
 
@@ -106,7 +107,17 @@ function MendelMiddleware(opts) {
         var pack = bpack({raw: true, hasExports: true});
         var decodedResults = trees.findTreeForHash(params.bundle, params.hash);
         if (!decodedResults || decodedResults.error) {
-            return notFound(res, decodedResults && decodedResults.error);
+            return bundleError(res, 404, decodedResults && decodedResults.error, params);
+        }
+
+        var modules = indexedDeps(decodedResults.deps.filter(Boolean));
+
+        if (!modules.length) {
+            // Something wrong, modules shouldn't be zero
+            return bundleError(res, 500, {
+                code: 'EMPTYBUNDLE',
+                message: 'Tree contents are empty'
+            }, params);
         }
 
         // Serve bundle
@@ -116,7 +127,6 @@ function MendelMiddleware(opts) {
         });
 
         pack.pipe(res);
-        var modules = indexedDeps(decodedResults.deps.filter(Boolean));
         for (var i = 0; i < modules.length; i++) {
             pack.write(modules[i]);
         }
@@ -277,14 +287,17 @@ function indexedDeps(mods) {
 
 ****/
 
-function notFound(res, error) {
+function bundleError(res, statusCode, error, params) {
     var message = "Mendel: ";
     if (!error) {
-        message += 'Bundle not found';
+        message += (statusCode === 404 ? 'Bundle not found' : 'Unable to create Bundle');
     } else {
         message += error.code + ' - ' + error.message;
     }
-    res.status(404).send(message);
+
+    debug([(error ? error.code : 'UNKNOWN'), 'hash=' + params.hash, 'bundle=' + params.bundle].join(' '));
+
+    res.status(statusCode).send(message);
 }
 
 function namedParams(keys, reqParams) {
--- a/packages/mendel-middleware/package.json
+++ b/packages/mendel-middleware/package.json
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "browser-pack": "^6.0.1",
+    "debug": "^2.6.3",
     "mendel-core": "^3.0.0",
     "mendel-loader": "^2.1.1",
     "path-to-regexp": "^1.2.1"


anuragdamle:
👍

packages/mendel-middleware/index.js
```diff
@@ -106,7 +106,17 @@ function MendelMiddleware(opts) {
...
+        }
+
+        var modules = indexedDeps(decodedResults.deps.filter(Boolean));
+
+        if (!modules.length) {
+            // Something wrong, modules shouldn't be zero
+            return notFound(res, {
+                code: 'EMPTYBUNDLE',
+                message: 'Tree contents are empty'
+            }, params);
```
irae: Humm, I don't think this is a good idea to do. This is only hiding the issue not solving it. There should be no way to get zero modules from the deps. If you fall into this condition, it is a bug somewhere that should not happen. This is uncovering something way more serious. There could be two things happening:

1. It could be a distributed problem, you might be getting `/mendel/*` requests on the old version of the app, while the new app is being deployed and the user got the root route `/` (the HTML) from the new box with the new hash. If this is the case, the fix is not here on the middleware.

2. You might have a bad manifest. Did you run the `validate-manifest` file against the manifests you are deploying? The script should work. It is inside `mendel-development` package but no one is requiring it. If the manifest is invalid, it might be causing the resolution to be non-deterministic and therefore create empty bundles only sometimes.

3. Somewhere in the code you might be mutating the manifest. It is kinda easy to know if this is happening, you can use `Object.freeze()` after reading the manifest. You can even do this using your Docker infrastructure and T-traffic production to this box. This way you test production without breaking production.

Either way, I don't think the fix is here. There are error codes for 404 and 500 in `mendel-core`. Maybe some of those are trowing and error is being swallowed?

packages/mendel-middleware/index.js
```diff
@@ -277,13 +286,20 @@ function indexedDeps(mods) {
...
+
+function notFound(res, error, params) {
     var message = "Mendel: ";
     if (!error) {
         message += 'Bundle not found';
     } else {
         message += error.code + ' - ' + error.message;
     }
+
+    logError(error ? error.code : 'UNKNOWN', params.hash, params.bundle);
```
irae: You have the message 2 lines above, why don't you log it?

packages/mendel-middleware/index.js
muralikr: I agree that this change is not fixing the issue. It will help us to find if this scenario is causing empty bundle issue. Since there are no Mendel logs in production and no way to debug this case on a pod, we need this change to catch this use case in production logs. Also, this will prevent sending empty bundle with 200 status code.

We have checked all available logs in when empty bundle was generated. We couldn't find related error messages or call stacks from mendel. It could be a bad manifest which is not uploaded properly to production. This change will prevent that scenario too.

As you mentioned, the actual issue still exists in our code. I want to prevent this issue and work on a solution. I don't want to keep the issue on production while I am working on fix.

packages/mendel-middleware/index.js
muralikr: I wanted to log errorCode along with hash and bundle. We may do more logging from other functions in this file. Hence, I moved it to a separate function.

packages/mendel-middleware/index.js
irae: Hey! I just realized something: You are applying `shasum` on the filename, right? You are not using `shasum` of the source code, correct? This might mean you have collisions even with `shasum`. When you have circular dependencies you can get zero entries added to the bundle because the infinite loop is prevented by the cache, but you never read the line that actually adds the bundle to the array.

Try removing shasum from `mendel-generator-prune` as explained in the other issue.

packages/mendel-middleware/index.js
```diff
@@ -277,13 +286,20 @@ function indexedDeps(mods) {
 
 ****/
 
-function notFound(res, error) {
+function logError(code, hash, bundle) {
+    console.log(['MendelError=' + code, 'hash=' + hash, 'bundle=' + bundle].join(' '));
```
irae: When you are logging errors it is better to `console.error` instead, this way it goes to error log and not access log. 

Please use `debug` module instead of creating this function, debug already uses `console.error` by default and `debug` module is industry standard and also used by `fluxible` and many teams at Yahoo.

You can then just set environment variables for Norrin in production to include `DEBUG="mendel*"` when you need it. You can also do `DEBUG="*error*` and you can get even `expressjs` errors, since `expressjs` also use `debug` internally.

packages/mendel-middleware/index.js
muralikr: `shasum` changes are not pushed to our prod instance. `shasum` change is increasing bundle size by huge margin. We had to get rid of `prune` to fix duplicate id assignment issue.

muralikr:
@irae Updated PR. Please take a look.

packages/mendel-middleware/index.js
irae: Getting rid of `prune` and getting rid on `shasum` and just having `normalizedId` is exactly the same. But when you remove prune you can accidentally include extra entire files. One more reason to fix `mendel-generator-prune` to remove the renaming index part.

packages/mendel-middleware/index.js
```diff
@@ -106,7 +107,17 @@ function MendelMiddleware(opts) {
         var pack = bpack({raw: true, hasExports: true});
         var decodedResults = trees.findTreeForHash(params.bundle, params.hash);
         if (!decodedResults || decodedResults.error) {
-            return notFound(res, decodedResults && decodedResults.error);
+            return throwBundleError(res, 404, decodedResults && decodedResults.error, params);
```
irae: This function don't need "throw" in the name because you don't `throw` on the function body. `bundleError` is enough.

irae:
👍  

[LGTM](https://lgtm.in/i/WpE8u9GSp)

packages/mendel-middleware/index.js
muralikr: fixed it.

PR #102, opened 2017-05-16
adding more options to the uglify compress and mangle sections

anuragdamle:
adding more options to the uglify compress and mangle sections. These options help ensure that we save some more bytes.

Options and the abilities are listed here
https://github.com/mishoo/UglifyJS2/tree/v2.x#compressor-options

@muralikr @irae plz review

### DIFF
--- a/packages/mendel-manifest-uglify/manifest-uglify.js
+++ b/packages/mendel-manifest-uglify/manifest-uglify.js
@@ -12,8 +12,8 @@ function manifestUglify(manifests, options, next) {
     // `compress` and `mangle` are set to `true` on uglifyify
     // just making sure we have the same defaults
     var uglifyOptions = Object.assign({
-        compress: true,
-        mangle: true,
+        compress: options.uglifyOptions.compress ? options.uglifyOptions.compress : true,
+        mangle: options.uglifyOptions.mangle ? options.uglifyOptions.mangle: true
     }, options.uglifyOptions, {
         fromString: true,
     });
--- a/packages/mendel-outlet-manifest/src/index.js
+++ b/packages/mendel-outlet-manifest/src/index.js
@@ -87,6 +87,8 @@ module.exports = class ManifestOutlet {
         manifestUglify([manifest], {
             uglifyOptions: {
                 root: this.config.baseConfig.dir,
+                compress: this.options.compress,
+                mangle: this.options.mangle
             },
         }, ([uglifiedManifest]) => {
             // happens immediately


irae:
You know you can pass all this in your `.mendelrc`, right? Or at least you should. If it doesn't bringing back `postProcessManifest` would load from config. I am not sure Stephan re-wronte the config part or dropped inadvertently.

When we created this plugin we followed the defaults of `grunt uglify` to be helpful with the community (by having same results), because Norrin already used defaults and because they have bugs filed against them because some unsafe options broke someone else's application. Not all of those are safe to every project. It really depends on how you structure your code, if you eslint (and wich rules), if you babel or other compiler.

I think it is best to only document on the readme for this package how to add to `.mendelrc`.

`mangle.toplevel` used to break Norrin in the past, specially lazy bundle integration. Did you try this on CI and acceptance tests passed? If so, this would be really great for Norrin, but as I said, this breaks many applications, I don't think it should be default for the middleware or for Mendel.

anuragdamle:
@irae i agree it should be as a part of config and values should come from .mendelrc. Having said that you mentioned that `mangle.toplevel` used to break stuff in norrin. Any specific things u recollect that would break. For us after running mendel in production mode, we are able to load lazy bundles without throwing any errors. Havn't tried running CI and acceptance test yet(will try soon) but manually checking of basic core flows doesnt seem to break anything.

irae:
I don't remember specifics, but yeah, it works for 90% of the application. The major problem with `mangle.toplevel` is that breaks developer confidence in the tool. You code perfectly, it works on dev, it passes on developer acceptance tests and once you push to CI and run against compiled code you get different and unintuitive results.

The only way to have `mangle.topLevel` without breaking developer flow is to have some kind of test (maybe linting) that make sure code is written in a way that it will compile correctly 100% of the time even with mangling. There are articles on that, I honestly don't recall how to do it. It is probably some couple of searches and reading away to be solved! If you find out let me know.

irae:
BTW, if you are so obsessed with payload, you will need to do the rollup clone implementation for Mendel we were planing for so long. It will be so much better than uglify because uglify will run on top of it. Also, you can run closureCompiler instead of uglify, with all the aggressive options code becomes even smaller. But you need to "code the right way" for it to work.

irae:
Not only this is more powerful (different projects with different settings) but commit looks way better too!

👍

muralikr:
👍

irae:
On this one we all forgot the bump

PR #104, opened 2017-05-16
Bring back postProcessManifest from accidental deletion

irae:
This commit only brings back post-process-manifest, whithout adding it to any packages that are currently in Mendel 3.x.
This might help us debug manifest problems and might be re-introduced in the bundle outlet steps, since internal data format hasn't change, using it will require small effort for a big pay-off.

Here is what this commit does:

  * Moves the post-process-manifest from old commit to `mendel-development`. It used to be on `mendel-cli`
  * Makes sure tests are updated and passing at least `test/post-process-manifest.js` passes well
  * Fix reference from `mendel-generator-extract` that used this formerlly, but tests are not passing because of `mendel-pipeline` integration on `test/generator-extract.js`, but at least fixes "Cannot find module" error

### DIFF
--- a/packages/mendel-development/post-process-manifest.js
+++ b/packages/mendel-development/post-process-manifest.js
@@ -1,0 +1,124 @@
+/* Copyright 2015, Yahoo Inc.
+   Copyrights licensed under the MIT License.
+   See the accompanying LICENSE file for terms. */
+
+var fs = require('fs');
+var path = require('path');
+var async = require('async');
+var inspect = require('util').inspect;
+var resolve = require('resolve');
+var debug = require('debug')('mendel-manifest-postprocess');
+
+var sortManifest = require('./sort-manifest');
+var validateManifest = require('./validate-manifest');
+
+module.exports = postProcessManifests;
+
+function postProcessManifests(config, finish) {
+    debug('start');
+    var resolveProcessor = processorResolver.bind(null, config);
+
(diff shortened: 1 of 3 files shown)


irae:
@muralikr @anuragdamle please review and just to be sure, this don't change anything on packages right now. I just did this recovery because having the file there is the first step to actually add it back to the codebase.

Just to give a bit of extra details, here is what I am proposing:

  1. Merge this PR, this changes almost nothing, just makes sure we can do other PRs that need this one firs.
  2. We can manually test using `postProcessManifest`. You can write a small file that runs this with hardcoded options against Norrin and see the results. Don't forget to run with `DEBUG=*` to see all messages, this file has a bunch of `debug` statements.
  3. Later we can replace [this two lines](https://github.com/yahoo/mendel/blob/328152fd40cbfc8a436be290ecf1cac27d13dca9/packages/mendel-outlet-manifest/src/index.js#L29-L30) by the post process manifest. This two lines are just hardcoded version of the `postProcessManifest`. The main difference is that `postProcessManifest` accepts how many plugins we need and knows how to read the config.

If we are going to do (3) or not is up for debate, but I think we could merge this one and keep iterating. I am just trying to help with recovering of this crucial piece on the mendel internal architecture.

muralikr:
👍

anuragdamle:
👍

PR #105, opened 2017-05-16
Add `falafel` as missing dependency of outlet server render

irae:
Here is how I found out this is needed:

  1. run `git clean -dfx`
  2. run `npm run linkall` from root
  3. run tests

A lot of things fail without falafel. I am pretty sure some people must have installed it by hand or even on their repo as dev-dependency because of this missing piece.

@anuragdamle @muralikr please take a look to review and merge as appropriate.


muralikr:
👍

PR #106, opened 2017-05-19
updating packages for uglify and outlet-manifest

anuragdamle:
@irae sorry for the PR just to bump versions. Will remember next time :)

@muralikr plz review

### DIFF
--- a/packages/mendel-development-middleware/package.json
+++ b/packages/mendel-development-middleware/package.json
@@ -14,9 +14,9 @@
   },
   "dependencies": {
     "path-to-regexp": "^1.7.0",
-    "mendel-config": "3.0.0",
+    "mendel-config": "^3.0.0",
     "mendel-development": "^1.1.0",
-    "mendel-exec": "3.0.0",
-    "mendel-pipeline": "3.0.0"
+    "mendel-exec": "^3.0.0",
+    "mendel-pipeline": "^3.0.0"
   }
 }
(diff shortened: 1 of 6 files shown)


muralikr:
👍

irae:
Are you sure you need to bump everyone? I agree you should bump to major all versions when the major came out. But I don't think you need `3.0.1` besides on uglify package and on prune package. All others could remain 3.0.1, right? Specially because those are unchanged.

packages/mendel-development-middleware/package.json
```diff
@@ -14,9 +14,9 @@
   },
   "dependencies": {
     "path-to-regexp": "^1.7.0",
-    "mendel-config": "3.0.0",
+    "mendel-config": "3.0.1",
     "mendel-development": "^1.1.0",
-    "mendel-exec": "3.0.0",
-    "mendel-pipeline": "3.0.0"
+    "mendel-exec": "3.0.1",
+    "mendel-pipeline": "3.0.1"
```
irae: I believe it should be `^3.0.0` actually this way we don't need to bump for minor versions of everything. 

Have you tried not bumping and just installing prune and uglify directly? If it does not work, can we fix by making this definition loose?

muralikr:
👍

irae:
Better! this way will reduce a lot the updates to `package.json` 👍

PR #107, opened 2017-05-23
ensuring the default fallback is set for uglify

anuragdamle:
referencing this fix https://github.com/yahoo/mendel/pull/102 we observed that the default options for compress and uglify are not being set to `true`. This PR will rectify the fallback regression.

@irae @muralikr plz review.


irae:
PRs are hard to review sometimes. Now that I see this diff I realize I was assuming this changes to be on outlet on the last PR but I didn't read the filename saying "manifest-uglify". Yeah, this new change set makes more sense. Feels responsibilities are on a better place. Sorry I didn't catch this last time.

Also, I am not checking out and testing most of the PRs recently. Just as FYI. I am doing code review and taking a look at the architecture mostly.

this is a good one to merge
👍

muralikr:
👍

PR #108, opened 2017-05-26
Better debugging scenarios for wrong requires or imports in client code

muralikr:
As of now, we are not throwing errors for unresolved file dependencies during `daemon` process. Errors are thrown while generating bundles in some cases. I observed some cases where `mendel` didn't throw any errors on invalid import paths.

This change throws errors during `daemon` process and provides clear message to developer. I am limiting this to file imports only.

Example error message:
`[Pipeline] "IndependentSourceTransform" errored:
Errored while resolving '../button' dependency for ./isomorphic/base/components/body.js`

@irae @anuragdamle

### DIFF
--- a/packages/mendel-exec/index.js
+++ b/packages/mendel-exec/index.js
@@ -136,6 +136,11 @@ module.exports = {
 
                     let normId = parent.deps[depLiteral][runtime];
                     if (typeof normId === 'object') normId = normId[runtime];
+
+                    // If we get _noop from cache, this depLiteral doesn't exist
+                    if (normId === '_noop')
+                        throw new Error(`Cannot find ${depLiteral} from ${mainEntry.id}`); // eslint-disable-line max-len
+
                     return resolve(normId);
                 },
             });
--- a/packages/mendel-pipeline/src/cache/index.js
+++ b/packages/mendel-pipeline/src/cache/index.js
@@ -319,7 +319,7 @@ class MendelCache extends EventEmitter {
 
             normDep[mod] = {};
             RUNTIME.forEach(runtime => {
-                let rtDep = dep[runtime];
+                let rtDep = dep && dep[runtime];
                 if (rtDep === false)
                     return normDep[mod][runtime] = false;
 


irae:
### About throwing

TL;DR: Avoid saying "throw" unless you use the `throw` statement, and daemon should never `throw`.

I'll address the PR in next comment, but first, lets make sure we are on same page as of vocabulary and the deamon process role. Reading the PR code it is clear to me that you don't `throw` errors. You just emmited the error back from the pipeline-step to the rest of the daemon. When you said `throw` it gave me the wrong impression, try saying "emitting errors" or "propagating errors" to better convey the issue.

Now, I am aware your PR is not breaking the design, but let's recap, because it will be important for the fix:

By design Daemon should never `throw` (in the sense of halting the process and exiting with error). It is meant to reach stability and eventually become something we can install on the mac that it can run as a launchd process, meaning, it restarts with the computer and does not need a terminal.

There are some reasons it is a bad idea to let the daemon `throw` and exit:

  * If you let the daemon throw and the process die, there is a high chance some of the workers will not exit cleanly and become a dangling process until user restarts.
  * If you have multiple bundles the error might be on a secondary bundle and the main bundle could still be saved
  * If the error is on a variation you are debugging and comparing with `base`, you don't want base bundle to stop being served.
  * In all the scenarios above, you would kill the daemon cache and this means slowly starting up again. The cache in memory is what makes Mendel super fast and reliable.

All that said, the problem of not finding the deps are already covered in other places of the code. Let me recap how it should work:

  1. Daemon parses all files and keep in memory all deps. Unresolved deps are also kept in memory as unresolved. It is not the job of the daemon or the transformer to fix this. It might be an error, or might be that an GST or outlet will yet do something. At this point, any `entry` is just information about a file, that had the sources transformed and deps detected. Including detected that it is missing.

  2. The cache is available ans served from daemon. Eventually a client connects and outlets are run and all the cache is fecthed. If your bundle has a file with a missing dependency, it should now render an invalid bundle. (you can try this on master on full example, it works). If an error is on the other hand in a variation or in a bundle that you only trigger with lazy loading, the error should only show when the code is actually executed. In my testing, I chagend full example `isomorphic/variations/partner_C/components/toolbar.jsx` to require `./missing_button` instead. When I hit localhost, it does not show me the error, when I hit localhost with `?variations=partner_C` it shows the error both in my browser and on the full-example development server. This behavior is correct.

The most important thing to understand here is that: Although you might run the daemon in the same "terminal", the daemon is designed to be on a different window and log almost nothing. It should log/error/warn only if `DEBUG="mendel*"` is being used and console.error() if it is an error in the daemon, not in the application code.

Errors in application code need to be surfaced in the client istead. Either in the middleware, or in the browser itself.

irae:
Now to the PR.

The current behavior: "When there are errors in dependencies in the application code, the error is only displayed to the developer when the bundle is requested".

What I think you want to do with the PR: "Developer gets a log on the terminal when he/she saves a file  with wrong typed dependencies".

This might be useful, I agree. Although there are multiple scenarios for that. For instance, if I am doing a large refactor, let's say I am moving a lot of files in my application, while I am in the midst of it, I know a lot of stuff is broken, so I might not even care for errors. The other scenario is when I pull and the changes had errors (maybe two people in parallel did different things in different PRs and the error only happens after merging both). In this case the error is more useful.

I am not pretty sure. I think mendel is doing the right thing already.

In case you want to have better logging of this errors, I think it should be on the client, not on the server. Once the client-cache receives an entry it can validate it.

I was investigating and found the place I would put the validation and guess what, there are some validation for the precise case already:

https://github.com/yahoo/mendel/blob/4dd69a1145e7880ab8946dad63017e2d74a66232/packages/mendel-pipeline/src/registry/client.js#L50-L58

I think you are onto something. The only difference is when to `warn` or to `error` and in which console. I believe what you want is a `warn` in the line we ad the `_noop` bundle. But not an error.

If I understood @stephanwlee correctly (by reading the code, didn't talk to him), the `_noop` is making sure neither the `daemon` nor the `client` exists because of "application code error". Mendel is still running and we don't need to stop the server (middleware) or the test suite for that. Once again, I think Mendel already handles it correctly.

irae:
So, I had a quick brainstorm with @stephanwlee and also dug a bit more into the code. Here are my personal conclusion:

1. We should not implement as is. Sorry I didn't have a clean answer the first time I reviewed the PR. But I acknowledge that better validation and errors are needed. I just should not be in the IST.
2. Development is fine: I don't think we need to have this on development console. Not on the daemon and on the client we already show in browser and in client console.
3. I think the only place this would be really helpful is on the `mendel-outlet-manifest`. This is meant for one time build and CI and we traverse all bundles there, so it should be possible to find all missing `deps` at this point, after transforms and after `defaultShims`.

Here is the good news: The `validateManifest` does just that and I already recovered it from the dead. We just need to use it on the `mendel-outlet-manifest` and it will likely be the fix you want for this issue.

@muralikr Can you take a look and validating this, either with `validateManifests` or with custom implmentation on the `mendel-outlet-manifest`?

muralikr:
@irae manifest is triggered only during production mode. I am seeing this issue in development which is not triggering manifest.

I have updated code and this error is thrown only on client process. Please review.

irae:
@muralikr I think we are not getting to a good place with this because you are mixing "Mendel issues" with "Code that is from your application that has issues". I think maybe my previous explanation about each piece of architecture might be badly explained or inaccurate, so let me try again:

Mendel has 2 groups of process, each designed to work differently, an lastly your application code:

1. Mendel Daemon: Designed to run headless, almost no console output besides "it is up" or "there is a critical error". Even with critical errors daemon should not necessarily restart and try to recover. Only exist if "Mendel Daemon own implementation is broken".

2. Mendel Clients: When I say "clients" I mean any consumer of pipeline/client. This includes development middleware, mocha runner, production outlets and maybe we will add more in the future. Clients are also made to never throw, but they can have more verbose warnings in development mode. They should never throw because it forces you to restart your server. In the case your own application and also the startup I work on, we have APIs running there, we have other responsibilities server-wise and those should not be impacted or restarted because of errors in the client code.

    2.1. The outlet to production (the manifest outlet) is a bit of a special case, because you are preparing to ship, so it should halt on any defective manifest.

3. Client code. In your case it is the yahoo application, in my case this is my startup UI. This code is not part of Mendel. Mendel only builds this code. Errors in this code should stop client side from working. When SSR has a bug in this code, the middleware should not stop either, it should just render the error instead.

The reason for this architecture is quite simple: Developer productivity and correctness. If your client-side code is broken, you should not break your server side (middleware code). If your code breaks 1% of your application, the remaining 99% should continue to work so you fix the problem.

Mendel is meant for large scale applications, that have more than one bundle (often between 3 and 4) and also multiple variations. If you have two tabs open and are checking out different variations, when an error that only affects a variation happens, the variation that is clean should just continue to work. Another scenario is a bundle that is broken, the other bundles should continue to work. If you are running acceptance tests against your dev box, the server should not shutdown, you should have the proper number of tests failing, all related to the code that broke, but all other tests should pass, otherwise you are on a bad debugging scenario that everything looks broken and it is only 1 line of code on 1 lazy bundle in 1 variation that is not even in production. It is also bad for context switching. All workflows break when you force a middleware to stop to give errors.

Not to mention that Webpack and Browserify both work as Mendel works right now. None of them stops everything if a dependency is missing.

What I am trying to do here is to get you to understand the goal of not shutting down the daemon and not shutting down the middleware/or/application server. We should be on the same page because this will help a lot in the future.

Next comment about suggestions.

packages/mendel-exec/index.js
```diff
@@ -136,6 +136,11 @@ module.exports = {
 
                     let normId = parent.deps[depLiteral][runtime];
                     if (typeof normId === 'object') normId = normId[runtime];
+
+                    // If we get _noop from cache, this depLiteral doesn't exist
+                    if (normId === '_noop')
+                        throw new Error(`Cannot find ${depLiteral} from ${mainEntry.id}`); // eslint-disable-line max-len
+
```
irae: As explained. Please don't throw. Use `console.error('WARN: ')` instead.

muralikr: @irae After this error, we have a [catch block](https://github.com/yahoo/mendel/pull/108/files/1c12bb9cf3015c9a62920a7a7f8498ab302df8f4#diff-74a2a21565400d96ccd47fe47532fadeR147) that is logging these errors using `console.error`. We also have another scenario in same code block where we [throw RangeError](https://github.com/yahoo/mendel/pull/108/files/1c12bb9cf3015c9a62920a7a7f8498ab302df8f4#diff-74a2a21565400d96ccd47fe47532fadeL135).

Since we have a catch, process doesn't exit for this particular error. It just logs them on console. On updating file, client resumes sync and other operations.

```
GET / 500 1098 - 25.549 ms
Error: Cannot find ./button1 from ./isomorphic/base/main.js
    at resolver (/Users/muralikr/dev/workspace/mendel/packages/mendel-exec/index.js:142:31)
    at varRequire (/Users/muralikr/dev/workspace/mendel/packages/mendel-exec/index.js:93:23)
    at ./isomorphic/base/components/body.js:25:16
    at runEntryInVM (/Users/muralikr/dev/workspace/mendel/packages/mendel-exec/index.js:30:9)
    at varRequire (/Users/muralikr/dev/workspace/mendel/packages/mendel-exec/index.js:95:20)
    at ./isomorphic/base/components/app.js:17:13
    at runEntryInVM (/Users/muralikr/dev/workspace/mendel/packages/mendel-exec/index.js:30:9)
    at varRequire (/Users/muralikr/dev/workspace/mendel/packages/mendel-exec/index.js:95:20)
    at ./isomorphic/base/main.js:3:12
    at runEntryInVM (/Users/muralikr/dev/workspace/mendel/packages/mendel-exec/index.js:30:9)
[Mendel] File change detected. Waiting to sync again...
[Mendel] Synced
```

I felt that current code is better since it shows proper error message with call stack. Please let me know if you want to change this to `console.error`. I am ok with either ways.

irae: OK, lets merge this for now.

I still think this is conceptually wrong and I'll open another issue so we eventually improve.

Call stacks are for developers of Mendel, those are client errors detected. I don't think we should be throwing for all this sort of client errors. We should be empowering developers to fix the errors without need to dive into Mendel code.

PR #109, opened 2017-05-31
Fix configuration for mocha-test to work

stephanwlee:
In the full-example, you can do `npm run daemon` and, in another tab, `npm run test`.

### DIFF
--- a/examples/full-example/.mendelrc
+++ b/examples/full-example/.mendelrc
@@ -72,7 +72,8 @@ types: # Type declaration of entries to be transformed or parsed
     transforms:
       - babel
     glob:
-      - "./**/_test_/*"
+      # Test files themselves should not be be instrumented.
+      - "**/_test_/*"
   less:
     extensions:
       - .less
@@ -175,8 +176,8 @@ env:
         transforms:
           - babel-test
         glob:
-          - "./**/*.{js,jsx}"
-          - "!./**/_test_/*"
+          - "**/*.{js,jsx}"
+          - "!**/_test_/*"
   production:
     types:
       js:
--- a/packages/mendel-pipeline/src/cache/entry.js
+++ b/packages/mendel-pipeline/src/cache/entry.js
@@ -1,4 +1,10 @@
 class Entry {
+    static getTypeForConfig(typeConfigs, id) {
+        if (isNodeModule(id)) return 'node_modules';
+        const type = typeConfigs.find(type => type.test(id));
+        return type ? type.name : '_others';
+    }
+
     constructor(id) {
         this.id = id;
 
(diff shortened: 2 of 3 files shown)


stephanwlee:
@anuragdamle @irae please take a look.

irae:
This is looking good to me 👍 
Thanks for the fix, @stephanwlee. Very much appreciated!

I don't know how we ended up here, looks like the `Entry` method was dead code with some duplication find logic on the cache. I wonder what are @stephanwlee's thoughts on having this be a cache member instead of static on entry though.

Verified that tests pass on `full-example` now. 

We need to bump the package version, but we can do this after merging the PR.

stephanwlee:
@irae RE:bumping - Yes, I prefer it to be incremental and hope the guys with bigger & more complex project to validate the fix before bumping.

anuragdamle:
so the fix works for the mocha-tests in the base folder. Just verified that.
screen shot 2017-05-31 at 11 12 37 am

Although the problem of tests not working for variations remains. We have a fix that works for base as well as variations, but wanted your opinion if this is ok to merge.

irae:
I think this is a good fix for now. This allows us to use the code transformed by mendel to test the code. I think in your project, @anuragdamle, you still do it with a custom test runner with babel only, right? This already enables you to move to `mendel-pipeline` providing the files.

Were you able to test it on the bigger project? Would be nice to know if you make it work. I used to have a branch with testing setup for this. It is quite easy, you just need to start a new `karma.conf` following this `full-example`.

muralikr:
@irae: @anuragdamle was referring to unit tests covered by this command. With this fix, `npm run test` in full-example will run unit tests from `base` variation only. We should cover unit tests from all variation folders.

We found that variation unit tests were not covered because of glob match in `registry`. I am okay to fix that in a follow up PR.

irae:
Yeah, the variations never worked. I created the variations test to have a target to aim for. We had karma-mendel in progress when we decided to stop and create mendel v2 (which became mendel 3).

muralikr:
For variations, @stephanwlee created [dependency resolver](https://github.com/yahoo/mendel/blob/master/packages/mendel-mocha-runner/index.js#L41-L68) for unit tests in `mendel-mocha-runner`. It is able to resolve dependencies from `base` and able to run those tests.

Do you think we have more unknowns in variation unit tests?

irae:
I think we can merge this one. I added some comments on the other one.

The `mendel-mocha-resolver` just need to be aware of the "current variation". The globs it is requesting do not include the variation. If if dit, it would not have triggered `isPadded`.

I think we are close to done on the tests. This fixes the `cache` and the `full-example` problem. We should just merge.

Moving forward we need:

1. Contextualized `getEntriedByGlob` on mocha runner
2. Finish writing `karma-mendel` so we have two options of runners.

Karma mendel should be easy to fix. We just need to push all files and dependencies very much like the mocha is doing, and have the resolver on the bridge. Each one of the items above can be separate PRs.

irae:
I don't know of any unknown after those 2 I mentioned.

PR #110, opened 2017-05-31
Fix some `npm` warnings

irae:
When running `npm run linkall` to install and cross-link for development, we get a log of warnings.

I solved all of them besides the ones in `full-example` that are related to sub-dependencies.

I tried to make this compatible with npm@5 but something is amiss, so I'll open a separate issue for that. Please test and try with npm@4.x.x for the time being.


muralikr:
👍

PR #111, opened 2017-05-31
mocha-test to work for variations

anuragdamle:
making mocha-tests work for variations.

### DIFF
--- a/packages/mendel-pipeline/src/registry/client.js
+++ b/packages/mendel-pipeline/src/registry/client.js
@@ -92,7 +92,8 @@ class MendelOutletRegistry {
 
             const isPadded = this._options.variationConfig.allVariationDirs
                 .some(varDir => str.indexOf(varDir) >= 0) ||
-                str.indexOf(this._options.baseConfig.dir) >= 0;
+                str.indexOf(this._options.baseConfig.dir) >= 0 ||
+                this._options.environment === 'test';
 
             if (!isPadded)
                 str = path.join(this._options.baseConfig.dir, str);
--- a/examples/full-example/.mendelrc
+++ b/examples/full-example/.mendelrc
@@ -72,7 +72,7 @@ types: # Type declaration of entries to be transformed or parsed
     transforms:
       - babel
     glob:
-      - "./**/_test_/*"
+      - "**/_test_/*"
   less:
     extensions:
       - .less
@@ -175,8 +175,8 @@ env:
         transforms:
           - babel-test
         glob:
-          - "./**/*.{js,jsx}"
-          - "!./**/_test_/*"
+          - "**/*.{js,jsx}"
+          - "!**/_test_/*"
   production:
     types:
       js:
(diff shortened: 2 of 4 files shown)


irae:
Your changes are based on PR#109 but you removed the following line:

`# Test files themselves should not be be instrumented.`

I think this is a good one to keep, it explains mendelrc better.

packages/mendel-pipeline/src/registry/client.js
```diff
@@ -92,7 +92,8 @@ class MendelOutletRegistry {
 
             const isPadded = this._options.variationConfig.allVariationDirs
                 .some(varDir => str.indexOf(varDir) >= 0) ||
-                str.indexOf(this._options.baseConfig.dir) >= 0;
+                str.indexOf(this._options.baseConfig.dir) >= 0 ||
+                this._options.environment === 'test';
```
irae: I don't think you want to remove the padding, this might cause all sorts of side effects. For instance, you would run this against all other folders of the project during test, including server only files and etc. You got lucky this actually works on full example.

Even if you do, the correct place is to change the if, not the `isPadded` boolean. `isPadded` is only about this particular string.

What I think we should be doing is to pad against all variations. Here is a draft I did quickly:

```diff
diff --git a/packages/mendel-pipeline/src/registry/client.js b/packages/mendel-pipeline/src/registry/client.js
index f7da91c..9b032d3 100644
--- a/packages/mendel-pipeline/src/registry/client.js
+++ b/packages/mendel-pipeline/src/registry/client.js
@@ -86,7 +86,7 @@ class MendelOutletRegistry {
 
     getEntriesByGlob(globStrings) {
         globStrings = Array.isArray(globStrings) ? globStrings : [globStrings];
-        const globs = globStrings.map(str => {
+        const globs = globStrings.reduce((resolvedGlobs, str) => {
             const isNegate = str[0] === '!';
             str = isNegate ? str.slice(1) : str;
 
@@ -94,14 +94,27 @@ class MendelOutletRegistry {
                 .some(varDir => str.indexOf(varDir) >= 0) ||
                 str.indexOf(this._options.baseConfig.dir) >= 0;
 
-            if (!isPadded)
-                str = path.join(this._options.baseConfig.dir, str);
-
-            if (!str.startsWith('./')) str = (isNegate ? '!./' : './') + str;
-            else str = (isNegate ? '!' : '') + str;
+            const globsToResolve = [];
+            if (isPadded) {
+                globsToResolve.push(str);
+            } else {
+                this._options.variationConfig.allVariationDirs
+                .forEach(varDir => {
+                    globsToResolve.push(path.join(varDir, str));
+                });
+            }
+
+            globsToResolve.forEach(str => {
+                if (!str.startsWith('./')) {
+                    str = (isNegate ? '!./' : './') + str;
+                } else {
+                    str = (isNegate ? '!' : '') + str;
+                }
+                resolvedGlobs.push(new Minimatch(str));
+            });
 
-            return new Minimatch(str);
-        });
+            return resolvedGlobs;
+        }, []);
 
         const positives = globs.filter(({negate}) => !negate);
         const negatives = globs.filter(({negate}) => negate);
```

This achieves the same thing. But this has potential to fix some other stuff, like the problem you had with lazy bundles not able to have a entrance file on `variations/foo` that don't exist on `base`. I do believe the above diff is the intended behavior for `getEntriesByGlob`. /cc @stephanwlee would you agree?

muralikr:
Closing this PR as it is not needed anymore. Similar fix is merged with https://github.com/yahoo/mendel/pull/114

PR #113, opened 2017-06-01
Make unix socket more resilient

stephanwlee:
This will require less need for manually removing unix socket file descriptor.

### DIFF
--- a/packages/mendel-pipeline/src/cache/network/unix-socket.js
+++ b/packages/mendel-pipeline/src/cache/network/unix-socket.js
@@ -1,0 +1,109 @@
+const BaseNetwork = require('./base-network');
+const net = require('net');
+const fs = require('fs');
+const {resolve} = require('path');
+
+function patchSocket(socket) {
+    const realWrite = net.Socket.prototype.write;
+    // Consistent API as WS or others this way.
+    const realEmit = net.Socket.prototype.emit;
+    const CONTENT_DELIMITER = '\u0004';
+    let buffer = '';
+    socket.send = function(str) {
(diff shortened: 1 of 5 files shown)


packages/mendel-pipeline/src/cache/network/unix-socket.js
```diff
@@ -0,0 +1,109 @@
...
+
+        return new Promise((resolve, reject) => {
+            const client = this.getClient({path});
+            client.unref();
+            client.once('connect', () => {
+                client.end();
+                reject();
+            });
+            client.once('error', resolve);
+        })
```
irae: OK, besides all the refactor that I think is interesting to put up a patter to make it work over network eventually, This is what I wanted most for stability and easy of use. Thanks!

packages/mendel-pipeline/src/cache/network/unix-socket.js
```diff
@@ -0,0 +1,109 @@
...
+        .then(() => fs.unlinkSync(path))
+        // If connection was made OR unlink was unsuccessful
+        // throw and exit -- cannot recover.
+        .catch(err => {
+            console.error([
+                'Server cannot start when another server is active.\n',
+                'If no server process is active, please remove or kill',
+                    `"${path}" manually.\n`,
+                'This is a symptom of server process exiting preemptively.',
+            ].join(' '));
```
irae: If you `\n` on every line, why don't you `.join('\n')`?

stephanwlee:
I won't make further changes to the PR. I can't merge anymore as my write access was taken off. Please take an action accordingly. Thanks.

PR #114, opened 2017-06-01
Fixed mendel-mocha variational test issue

stephanwlee:
Bug in getEntriesByGlob filtered out the entries in the
variational directories.

### DIFF
--- a/packages/mendel-pipeline/src/registry/client.js
+++ b/packages/mendel-pipeline/src/registry/client.js
@@ -89,18 +89,12 @@ class MendelOutletRegistry {
         const globs = globStrings.map(str => {
             const isNegate = str[0] === '!';
             str = isNegate ? str.slice(1) : str;
+            // Strip "./" in case it is prepended already.
+            str = str.startsWith('./') ? str.slice(2) : str;
 
-            const isPadded = this._options.variationConfig.allVariationDirs
-                .some(varDir => str.indexOf(varDir) >= 0) ||
-                str.indexOf(this._options.baseConfig.dir) >= 0;
-
-            if (!isPadded)
-                str = path.join(this._options.baseConfig.dir, str);
-
-            if (!str.startsWith('./')) str = (isNegate ? '!./' : './') + str;
-            else str = (isNegate ? '!' : '') + str;
-
-            return new Minimatch(str);
+            // Have to prepend "./" because entry Ids starts with "./".
+            const pattern = `${isNegate ? '!' : ''}./**/${str}`;
+            return new Minimatch(pattern);
         });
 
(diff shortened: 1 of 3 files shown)


packages/mendel-pipeline/src/registry/client.js
```diff
@@ -110,7 +104,11 @@ class MendelOutletRegistry {
             return positives.some(g => g.match(id)) &&
                 negatives.every(g => g.match(id));
         })
-        .map(id => this.getEntry(id));
+        .map(id => this.getEntry(id))
+        .filter(({variation}) => {
+            return !excludeVariational ||
+                variation === this._options.baseConfig.dir;
+        });
```
irae: Why do we exclude variational in here? For instance, what about a variation that introduces a bundle? The _entrance_ would only exist in the new bundle. I was under the impression that picking up from all variation dirs was the proper behavior. It might be breaking something, but Not in full example, at least.

packages/mendel-mocha-runner/index.js
```diff
@@ -36,7 +36,7 @@ function MendelRunner(filePaths, options={}) {
 
         const mocha = new Mocha(options);
 
-        const entries = client.registry.getEntriesByGlob(filePaths);
+        const entries = client.registry.getEntriesByGlob(filePaths, false);
```
irae: See suggestion about not filtering. But if we keep the filtering, can we have meaningful options instead, even if this is the only config so far?

```js
// usage
client.registry.getEntriesByGlob(filePaths,{includeVariations:true});
// implementation
getEntriesByGlob(globStrings, options = {includeVariations:false}) {/*...*/}
```

packages/mendel-pipeline/src/registry/client.js
stephanwlee: >  I was under the impression that picking up from all variation dirs was the proper behavior. It might be breaking something, but Not in full example, at least.

Not true. 
https://github.com/yahoo/mendel/blob/master/examples/full-example/.mendelrc#L135
https://github.com/yahoo/mendel/blob/e46e7bf1c67d28c12f7566432e58a4406bbb9864/packages/mendel-pipeline/src/bundles/default-generator.js#L10

Imposing that you create a file in base variation for "entries" is not a bad one. Btw. `walk` will take care of variations. 
https://github.com/yahoo/mendel/blob/e46e7bf1c67d28c12f7566432e58a4406bbb9864/packages/mendel-pipeline/src/registry/client.js#L124

In summary, there may be no harm in walking in variational dir too but there is a strong chance it will do redundant walking.

stephanwlee: Will remove the filter :\

PR #117, opened 2017-06-02
Improved example of "support" field. 

stephanwlee:
full-example/package.json now includes different ways to
bootstrap test environment.

`npm run test` will bootstrap from "setup.js" whereas
`npm run test-dev` will bootstrap from glob
(all files - test/setup.js,test/setup/jsdom.js,test/setup/mocha-global.js)

### DIFF
--- a/examples/full-example/package.json
+++ b/examples/full-example/package.json
@@ -4,9 +4,9 @@
   "description": "Example of app using mendel, es6, jsx, react, server-side-rendering, multivariant and multilayer support.",
   "main": "index.js",
   "scripts": {
-    "test": "mendel-mocha --prelude test/setup/jsdom.js **/_test_/*.js",
-    "test-dev": "mendel-mocha --prelude test/setup/jsdom.js **/_test_/*.js --watch",
-    "coverage": "mendel-mocha --prelude test/setup/jsdom.js **/_test_/*.js --reporter=mocha-istanbul-reporter --reporter-options=text,text-summary",
+    "test": "mendel-mocha --prelude 'test/setup.js' **/_test_/*.js",
+    "test-dev": "mendel-mocha --prelude 'test/setup/*.js' **/_test_/*.js --watch",
+    "coverage": "mendel-mocha --prelude 'test/setup/*.js' **/_test_/*.js --reporter=mocha-istanbul-reporter --reporter-options=text,text-summary",
     "build": "NODE_ENV=production mendel",
     "daemon": "mendel --watch",
     "development": "nodemon index.js",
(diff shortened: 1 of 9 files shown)


examples/full-example/package.json
```diff
@@ -4,9 +4,9 @@
   "description": "Example of app using mendel, es6, jsx, react, server-side-rendering, multivariant and multilayer support.",
   "main": "index.js",
   "scripts": {
-    "test": "mendel-mocha --prelude test/setup/jsdom.js **/_test_/*.js",
-    "test-dev": "mendel-mocha --prelude test/setup/jsdom.js **/_test_/*.js --watch",
-    "coverage": "mendel-mocha --prelude test/setup/jsdom.js **/_test_/*.js --reporter=mocha-istanbul-reporter --reporter-options=text,text-summary",
+    "test": "mendel-mocha --prelude 'test/setup.js' **/_test_/*.js",
+    "test-dev": "mendel-mocha --prelude 'test/setup/*.js' **/_test_/*.js --watch",
```
stephanwlee: For glob in prelude, you must surround it with quotation mark. Otherwise, bash will expand it to files.

stephanwlee:
@anuragdamle note that `support` is not strictly needed for mendel-mocha-runner. If we cannot find preludes in the `support` or cache, we [evaluate the code out of text](https://github.com/yahoo/mendel/blob/master/packages/mendel-mocha-runner/index.js#L46).

This might be favorable if your prelude (like lodash?) brings in a huge dependency which otherwise would not be there. Huge dependency chain can cause slower daemon as we need to transform more files and what not. Btw, because we evaluate the code out of text, it is very preferable to have that piece of code written in Es6 *without* ES Module (unless Node8+ supports it in the future).

packages/mendel-generator-extract/generator-extract.js
```diff
@@ -33,6 +33,7 @@ function generatorExtract(bundle, doneBundles, registry) {
     // Collect dependencies of main as if lazy was not there
     registry.getEntriesByGlob(fromBundle.options.entries)
     .forEach(({normalizedId, type}) => {
+        console.log(normalizedId, type)
```
irae: woops?

packages/mendel-pipeline/src/daemon.js
```diff
@@ -163,6 +163,12 @@ module.exports = class MendelPipelineDaemon {
 
             this.cacheManager.sync(cache);
             this.pipelines[environment].watch();
+
+            // `config.support` is a special configuration
+            // that can dynamically add more entries to the pipeline.
+            if (envConf.support) {
+                this.watcher.subscribe([envConf.support]);
+            }
```
irae: Just wondering: Can we consider "support" a way to require something in all bundles? based on type? I don't fully understand the concept. Maybe add this to some readme? Do we have a readme for mendelrc anyway?

muralikr:
👍

irae:
👍

PR #118, opened 2017-06-02
Parallelism tweak

stephanwlee:
Playing with child_process.

After 4, there is a diminishing marginal utility at cost of memory. So we cap at 4 with minimum of 2.

### DIFF
--- a/packages/mendel-pipeline/src/multi-process/base-master.js
+++ b/packages/mendel-pipeline/src/multi-process/base-master.js
@@ -2,7 +2,7 @@ const analyticsCollector = require('../helpers/analytics/analytics-collector');
 const analyzeIpc = require('../helpers/analytics/analytics')('ipc');
 const {fork} = require('child_process');
 // After 4, there is a diminishing marginal utility at cost of memory.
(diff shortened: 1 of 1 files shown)


packages/mendel-pipeline/src/multi-process/base-master.js
```diff
@@ -1,7 +1,8 @@
 const analyticsCollector = require('../helpers/analytics/analytics-collector');
 const analyzeIpc = require('../helpers/analytics/analytics')('ipc');
 const {fork} = require('child_process');
-const numCPUs = require('os').cpus().length;
+// After 4, there is a diminishing marginal utility at cost of memory.
+const RECOMMENDED_CPUS = Math.min(2, Math.max(4, require('os').cpus().length));
```
muralikr: So, this will be always less than or equal to 2 ?? Won't that make daemon much slower??

stephanwlee: Yeah I flipped by mistake. Will fix it.

muralikr:
👍

irae:
Nice!
👍

PR #119, opened 2017-06-02
Allow usage of `global` in isomorphic code

stephanwlee:
@anuragdamle @irae 

By the way, the shims are not being used in any meaningful way as of right now.

Details:
If you see how we browser-pack, we did a quick hack to detect that a bundle requires a global like "process" (they are special in a sense that there won't be any "require" associated with it) and inject custom code. The custom code is the hacky part. Instead of requiring those shim like `(function(){process=require('process');...`, we effectively hand-written the same shims because method `require` is not defined yet and we need to have them before browser-pack's require function invokes the entries.

https://github.com/substack/browser-pack/blob/master/prelude.js#L40

We may be able to muck around with entry by prepending or passing custom prelude.js.

### DIFF
--- a/packages/mendel-config/src/defaults.js
+++ b/packages/mendel-config/src/defaults.js
@@ -1,6 +1,7 @@
 /* Copyright 2015, Yahoo Inc.
    Copyrights licensed under the MIT License.
    See the accompanying LICENSE file for terms. */
+const path = require('path');
 
 module.exports = function() {
     const mendelEnv = process.env.MENDEL_ENV ||
@@ -42,8 +43,17 @@ module.exports = function() {
         env: {},
         bundles: {},
         config: true,
-        shim: {},
-        defaultShim: {},
+        shim: {
+            // Can pass any shim that would resolve modules differently.
+            // For instance, you can pass "fs" and name of the package
+            // to inject different implementation of "fs" instead of that of
+            // the browser.
+            // This would override "defaultShim" below.
+        },
+        defaultShim: {
+            // For Mendelv2, the default set of shim is listed in
+            // https://github.com/webpack/node-libs-browser.
+        },
         ignores: [],
         // This controls whether outlet outputs to a file or a stream
         noout: false, // TODO re-evaluate whether we need this guy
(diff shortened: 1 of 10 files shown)


examples/full-example/isomorphic/base/components/app.js
```diff
@@ -8,6 +8,10 @@ import Body from './body';
 import Footer from './footer';
 import './about';
 
+global.foo = function() {
+    console.log(process.env.NODE_ENV);
+};
+
```
irae: We could use this function in the footer, no? To show environment 😄

packages/mendel-outlet-browser-pack/src/index.js
```diff
@@ -65,10 +72,12 @@ module.exports = class BrowserPackOutlet {
...
             let appendix = '';
 
-            if (hasProcess) {
-                prelude = '(function(){var process={env: {}};';
+            if (hasProcess || hasGlobal) {
+                prelude = '(function(){';
                 appendix = '})();';
             }
+            if (hasProcess) prelude += 'var process={env: {}};';
+            if (hasGlobal) prelude += 'var global=window;';
```
irae: shouldn't this be something like:

```js
if (hasGlobal || hasProcess) prelude += 'var global=window;';
if (hasProcess) prelude += 'var process=global.process||{env: {}};';
```
This would be a better polyfill for process because process is the same instance across all modules.

packages/mendel-config/src/defaults.js
```diff
@@ -42,7 +43,9 @@ module.exports = function() {
         env: {},
         bundles: {},
         config: true,
-        shim: {},
+        shim: {
+            global: path.join(__dirname, 'default-shims', 'global.js'),
+        },
```
irae: I don't see this being used. Isn't global shim hardcoded in outlet?

packages/mendel-outlet-browser-pack/src/index.js
irae: I can merge first and change later if you are time constrained today but you agree it is a good idea.

irae:
👍 

@muralikr @anuragdamle would be good to have your feedback on this. I think you are shimming global manually, right?

PR #121, opened 2017-06-03
Prevent requiring in transformer in Mendel-config

stephanwlee:
Because of "require" (another option is to use AST parser),
for instance of transform-babel, we required all of babel which
contributed to large bootstrap time on low-powered computer.

### DIFF
--- a/packages/mendel-config/src/transform-config.js
+++ b/packages/mendel-config/src/transform-config.js
@@ -1,18 +1,32 @@
 var createValidator = require('./validator');
 var nodeResolveSync = require('resolve').sync;
+var path = require('path');
+
+function _resolve(pluginPath, opt) {
+    try {
+        return nodeResolveSync(pluginPath, opt);
+    } catch (e) {
+        return false;
+    }
+}
+
 
 function TransformConfig(id, transform, {projectRoot}) {
     this.id = id;
+    const basedir = projectRoot;
+    const pluginPackagePath = _resolve(path.join(transform.plugin, 'package.json'), {basedir});
+    const pluginPath = _resolve(transform.plugin, {basedir});
 
-    try {
-        this.plugin = nodeResolveSync(transform.plugin, {basedir: projectRoot});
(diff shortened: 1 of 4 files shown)


irae:
Although startup is much faster. This breaks development for me. You get bundles without transformation. JSX in my browser throwing syntax error for `<`.

stephanwlee:
@irae that is not intended behavior of this change. This should only affect config. Can you show me the config thru slack? Thanks

stephanwlee:
I have no idea how to reproduce the problem you noticed. Can you please commit a example with a config that breaks this change in the upstream so we can test that? Otherwise, I am just merging it.

irae:
Please bump the appropriate packages before merging.
And: Oh yeah, we can merge, I didn't merge because you said you are not ready.

The issues I had were probably caused, in order of likelihood:

1. `mendel-transform-buble` to be missing stuff in `package.json`
2. I had a mix of npm installed and npm linked packages
3. npm acting weird and missing dependencies and/or mixing versions
4. running my global installation of Mendel instead of local installation on my repo.

packages/mendel-config/src/transform-config.js
```diff
@@ -1,18 +1,32 @@
...
-        this.plugin = nodeResolveSync(transform.plugin, {basedir: projectRoot});
-        this.mode = require(this.plugin).mode || 'ist';
-    } catch (e) {
-        if (e.code !== 'MODULE_NOT_FOUND') throw e;
-        this.plugin = false;
+    if (pluginPath) {
+        this.mode = [pluginPackagePath, pluginPath]
+            .filter(Boolean)
+            .map(plugin => require(plugin).mode)
+            .filter(Boolean)[0] || 'ist';
```
irae: I don't think you are preventing the require. You would prevent if you used `.some`, right?
The map always goes through all resolved. If both plugin and package.json resolves, you require both, no?

Maybe I am missing something.

irae:
closing in favor of #126

PR #123, opened 2017-06-04
Small fix to accept applications without ssr too

irae:
I was testing mendel with a project without SSR and it throws when trying to create `ssrBundle` without any config bundles.

The rest of the code seems to have some checks without SSR, so this is the only small fix we need to support client-side only repos.


stephanwlee:
👍 Interesting to see a bug with mendel-core :)

muralikr:
👍

PR #126, opened 2017-06-06
Avoid config requires

irae:
This PR fixes two issues:

1. Config itself never uses the transforms, so `require`ing it is expensive and memory consuming for no reason. Without this we have faster startup for daemon in development mode
2. In production, if you `npm prune --production` you can't require the transforms since they are dev dependencies
3. When you first install mendel, it is tedious to fix one by one of the packages you might have forgotten, so this will also log to stdrr instead of throwing on the first not found package.

### DIFF
--- a/packages/mendel-config/src/resolve-plugin.js
+++ b/packages/mendel-config/src/resolve-plugin.js
@@ -1,0 +1,47 @@
+var nodeResolveSync = require('resolve').sync;
+var path = require('path');
+
+function resolvePlugin(pluginName, basedir) {
+    const pluginPackagePath = _resolve(path.join(pluginName, 'package.json'), {
+        basedir,
+    });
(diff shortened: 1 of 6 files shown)


packages/mendel-config/src/resolve-plugin.js
```diff
@@ -0,0 +1,47 @@
...
+                // ¯\_(ツ)_/¯
+            }
+        }
+        return mode;
+    }, false);
+
+    resolved.mode = resolved.mode || 'unknown';
+
+    if (resolved.mode === 'unknown' && process.env.NODE_ENV !== 'production') {
+        console.error(
```
stephanwlee: don't feel too hot about this.

irae: propose something instead?

Issue #86, opened 2017-03-07
Manifest validation needed


irae:
Making a manifest is hard enough, without validation is super hard for someone not familiar with mendel internals.

For instance, there is a "outlet" requirement on the "bundle". If you don't match exactly with an existing  outlet plugin by `id` you will get an error deep down on inside the compiler chain. Instead we should warn at boot of the pipeline.

Al required fields should be there, al folders should exist, all references by ID should validate and all modules should be loaded successfully before we start mendel. I think having validation will pay off in the long run better than having a installation script, since people start projects simple and add stuff as they progress.

irae:
Very related to #82

irae:
Ha! I forgot I already knew about this.

I believe the way to go is to recover `mendel-postprocess-manifest`. It used to be a file inside `mendel-browserify` and the sorting and validation is still unused files on `mendel-development-middleware`.

It should be easy to make `mendel-outlet-manifest` incorporate the architecture and we could instead of forcing to uglify have this done per configuration, opening the architecture to have people use `babily` if they so prefer instead of uglify.

Either way, it should be easy to have sorting and validation added back because the data structures are still the same. `mendel-manifest-uglify` being used without changes proves this is the case.

irae:
closed by #130

irae:
I reverted PR #130 and we will be working on this again. No harm done as I didn't push this versions to npm yet.

irae:
Older validation was breaking because we don't support the notion of "external" on the mendel-pipeline right now.

We do support:
  * "expose": when a module is included on a bundle but made available to other bundles by a given name

We do not support:
  * "external": When I modules is not on a given bundle, but we expect it to be available on a different bundle.

This has the same underlying issue we saw when @muralikr and @anuragdamle decided to shasum in `mendel-generator-prune` and later decided they would not use "prune" anymore. The reason was: `mendel-generator-prune` didn't have enough information to avoid compressing IDs on bundles that had external sources.

In "browser-pack" or in browserify land we used to mark externals with "boolean false" which is different from undefined or something else:

https://github.com/yahoo/mendel/blob/aca231e37b1f56fed5dd14a5ec969eb7830868b8/packages/mendel-manifest-extract-bundles/manifest-extract.js#L102

This was easy to miss when moving to mendel v3. I think it is likelly we need to have our own way of conveying "externality" of deps.

irae:
And the validation counterpart:

[...]

PR #130, opened 2017-06-08
Add sorting and validation to all manifests in mendel-outlet-manifest

irae:
Mendel 2 is way better in terms of consistency. Simulating the validation errors is quite hard, but might happen in case of scenarios like missing files and deps. Validation is still useful in this scenarios since it forces production build to break when generating bundle manifests.

Sorting comes handy in the long term. Different teams might have different workflows. Some can build and commit manifest to git, others can package production builds with private instance of npm and others might have docker containers. What does not change is the ability to compare manifests between different production builds. Sorting becomes very valuable in this case, since generating a diff of changes is quite useful for debugging reasons, for bundle analysis and for debugging issues across mendel versions (although Mendel is quite stable now, it might happen in the future to beta users of next major versions).

This PR helps in both fronts by bringing back sort and validation functions that are quite stable and proven to be useful since Mendel v1.

### DIFF
--- a/packages/mendel-development/package.json
+++ b/packages/mendel-development/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mendel-development",
-  "version": "3.0.0",
+  "version": "3.0.1",
   "description": "Mendel shared dependencies for build and development",
   "main": "index.js",
   "scripts": {
--- a/packages/mendel-outlet-manifest/package.json
+++ b/packages/mendel-outlet-manifest/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mendel-outlet-manifest",
-  "version": "3.0.3",
+  "version": "3.1.0",
   "description": "Mendel outlet for generating mendel-core (Mendel v1) compatible manifest. Manifest is a serialization of a Mendel cache.",
   "main": "src/index.js",
   "scripts": {
@@ -17,6 +17,7 @@
     "babel-plugin-transform-inline-environment-variables": "0.0.2",
     "debug": "^2.6.0",
     "mendel-manifest-uglify": "^3.0.2",
+    "mendel-development": "^3.0.1",
     "shasum": "^1.0.2"
   }
 }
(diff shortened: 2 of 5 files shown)


packages/mendel-development/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mendel-development",
-  "version": "3.0.0",
+  "version": "3.0.1",
```
stephanwlee: What do you think about having one Mendel version? Akin to react and barbel.

packages/mendel-outlet-manifest/package.json
```diff
@@ -17,6 +17,7 @@
     "babel-plugin-transform-inline-environment-variables": "0.0.2",
     "debug": "^2.6.0",
     "mendel-manifest-uglify": "^3.0.2",
+    "mendel-development": "^3.0.1",
```
stephanwlee: Not sure about this dependency. What else does Mendel development do? From the name of the package, it is support hard to know.

stephanwlee: Can we just have mendel-manifest-util?

irae: I think moving this to outlet is the future outcome. I just don't want o go through all the refactor right now because of time constraints. `mendel-development` is the grab bag for all cross package utilities since mendel-v1. If more than 1 package uses it. it is there. You can think of it as `mendel-developmente-lib` or `mendel-development-utils`. The other way to fix it is to move a lot of mendel-pipeline inside of mendel-development. 😄

packages/mendel-development/package.json
irae: `lerna` would have all this options. Just following old ways for now.
I don't like bumping all for a number of reasons, I tried that in the past and it is quite awkward, specially when you have lots of plugins that have different speeds of development. We can adopt this for all of core packages. It would be quite nice. But than it becomes a good idea to bake the changes in master before releasing a bit more slowlly

irae:
Closes #86

PR #132, opened 2017-06-09
Support for node globals in production by using mendel-browser-pack

stephanwlee:
Factored out the mendel-browser-pack out of mendel-outlet-browser-pack which handles our outlet specific treatment for node globals.

Production should be fixed with this.

### DIFF
--- a/packages/mendel-browser-pack/index-deps.js
+++ b/packages/mendel-browser-pack/index-deps.js
@@ -1,0 +1,106 @@
+/**
+ * Modules internal to a bundle can be referenced by index instead of id.
+ * Remap ids and deps in such case.
+ * @example
+ * This compresses the bundle by renaming all dependency indexes from file paths
+ * to a numbered index.
+ *
+ * Here is a sample transformation:
+ * ```js
+ * [
+ *   {
+ *     "entry": true,
+ *     "id": "/User/me/projects/site/src/main.js",
+ *     "deps": {
+ *       "./colors.js": "/User/me/projects/site/src/colors.js",
+ *       "./shared.js": "/User/me/projects/site/src/shared.js"
+ *     }
+ *   },
+ *   {
+ *     "id": "/User/me/projects/site/src/colors.js",
+ *     "deps": {
+ *       "external-lib": false
+ *     }
+ *   },
+ *   {
+ *     "expose": "shared",
+ *     "id": "/User/me/projects/site/src/shared.js",
+ *     "deps": {}
+ *   }
+ * ]
+ * ```
+ *
+ * Should become:
+ * ```js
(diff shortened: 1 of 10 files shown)


packages/mendel-middleware/index.js
```diff
@@ -126,117 +125,11 @@ function MendelMiddleware(opts) {
...
-            else {
-                // for all other props we just copy over
-                newMod[prop] = oldMod[prop];
-            }
-
-            return newMod;
-        }, {});
-    });
-}
-
```
irae: Nice idea to extract this into a module

stephanwlee:
Oh wait a minute though. I am going to make mendel-browser-pack pipeable like browser-pack.

packages/mendel-browser-pack/index-deps.js
```diff
@@ -0,0 +1,106 @@
+/**
+ * Modules internal to a bundle can be referenced by index instead of id.
+ * Remap ids and deps in such case.
+ * @example
+ * This compresses the bundle by renaming all dependency indexes from file paths
+ * to a numbered index.
+ *
+ * Here is a sample transformation:
+ * ```js
```
stephanwlee: this is from mendel-middleware, just fyi ;)

irae: I know, but this is also from when Mendel didn't even had a notion of generators.

muralikr:
Would have been great if we have a closing comment.

irae:
He closed this because of the architecture discussions. Global is a Common JS concept and we thought we might move this detection to the daemon process instead of the production middleware.

Yes, this is a small piece of code and very small overhead. But the fact that we have this implementation in `mendel-middleware` as well as in `mendel-development-middleware` is a bit of a smell. He closed the PR so to save time of people reviewing something we thought needed a bit more thought.

I think we have two options here to make this clean. In order of easy to implement:

1. We could create generic classes and move the "PaddedStream" to somewhere we can share between production and development middleware. This way we at least share code. We can have the code as generic enough so we can re-use later if new places need the same shim, for instance, maybe mocha runner needs it.

2. The better for architecture is if we could run detection at the daemon process and all clients would benefit of transparently getting this shim. That's what we decided to aim for. But to do that, we need to improve the "entry" type system on the daemon and on `.mendelrc`. We brainstormed about this on the Mendel Slack channel and over a couple of voice calls. This is the end goal architecture-wise, but it will be some work. The good part about this, is that it ties with a bit of simplification on `.mendelrc`, for instance, `parserToType` might be removed and configuration might become smaller.

Currently, the recommendation is to include a shim in your application, like I believe you are doing, Murali, IIRC.

